PR #20, opened 2015-02-18
Update template.jag

Vishanth:
Single line comment in mediation script is causing error when parsing synapse api artifact


Vishanth:
Thank you.

PR #30, opened 2015-02-27
Pull request for modified integration module compatible with git

dharshanaw:


modules/integration/pom.xml
```diff
@@ -17,10 +17,10 @@
 
     <modules>
         <module>backend-service</module>
```
abimarank: Another backend-service module was added inside tests-common module with same artifact ID. Please have a look.

PR #43, opened 2015-03-07
test cases for APIMANAGER-3359

chamilaadhi:
test cases for APIMANAGER-3359

### DIFF
--- a/modules/integration/tests-ui/src/test/resources/testng-server-mgt.xml
+++ b/modules/integration/tests-ui/src/test/resources/testng-server-mgt.xml
@@ -24,6 +24,7 @@
             <!--<class name="org.wso2.carbon.am.integration.ui.APIMANAGER2957EpochTimeTestCase"/>
             <class name="org.wso2.carbon.am.integration.ui.APIMANAGER3017ExternalStoreLifeCycleTestCase"/>
             <class name="org.wso2.carbon.am.integration.ui.APIMANAGER3149SSOLogin"/>-->
+            <!--<class name="org.wso2.carbon.am.integration.ui.APIResourceBodyParameterRemovalTestCase"/>-->
         </classes>
     </test>
 </suite>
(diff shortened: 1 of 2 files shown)


modules/integration/tests-ui/src/test/resources/testng-server-mgt.xml
```diff
@@ -24,6 +24,7 @@
             <!--<class name="org.wso2.carbon.am.integration.ui.APIMANAGER2957EpochTimeTestCase"/>
             <class name="org.wso2.carbon.am.integration.ui.APIMANAGER3017ExternalStoreLifeCycleTestCase"/>
             <class name="org.wso2.carbon.am.integration.ui.APIMANAGER3149SSOLogin"/>-->
+            <!--<class name="org.wso2.carbon.am.integration.ui.APIResourceBodyParameterRemovalTestCase"/>-->
```
krishanthasamaraweera: Why this test case need to be commented out ? any reason

chamilaadhi: I disabled that test case because all the other UI test cases are disabled. Had a chat with a team member for the reason. He mentioned that builds getting failed if we enable the UI tests

chamilaadhi: Since these are selanium test cases and selanium is not properly configured in the build machine these test cases failed. Need to configure the builder machine and enable these test cases.

krishanthasamaraweera: In that case we need to skip the tests from local build and enable tests execution on Jenkins..We are woking on Jenkins part. 

To skip the test from regular build you can follow. 

Add the following property to surefire configuration

<skipTests>${skipUiTests}</skipTests>

Then add following property to ui module pom.

 <properties>

        <skipUiTests>true</skipUiTests>

    </properties>

To run the UI tests, you need to use following command.

mvn install -DskipUiTests=false

PR #62, opened 2015-03-20
pt-BR i18n

sergerehem:
This is a contribution for wso2 to internationalize store and publisher webapps.

I could not test in development mode and I didn't have time yet to check if all entries are presented in both EN and PT-BR versions.

I think we need Brazilian Portuguese internationalization to increase wso2 adoption in portuguese spoken language countries, so I hope that you accept, test and evolve this modest contribution.


hevayo:
Thanks a lot for the contribution we will add this to the APIM.

But we wount be able to merge it directly since the store and publisher apps will get added to the build from the features section of wso2/carbon-apimgt.

Hopefully we can include this with the next release 1.9.

harsha89:
Closing as this is no longer valid for C5

bhathiya:
Could you please send this to 6.1.x branch? We can then merge it.

PR #147, opened 2015-05-07
Moving api-store-extension to product/src/main/ directory

thushara35:


modules/distribution/product/src/main/api-store-extension/am_store/app.js
```diff
@@ -0,0 +1,27 @@
+/*
+ *  Copyright (c) 2005-2014, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
cnapagoda: For new file, year should be 2015

PR #153, opened 2015-05-08
Adding publisher landing page and extensions

thushara35:


modules/distribution/product/src/main/api-publisher-extension/am-publisher-common/app.js
```diff
@@ -0,0 +1,28 @@
+/*
+ *  Copyright (c) 2005-2014, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
cnapagoda: For new file, year should be 2015

modules/distribution/product/src/main/api-publisher-extension/am-publisher-defaults/app.js
```diff
@@ -0,0 +1,19 @@
+/*
+ *  Copyright (c) 2005-2014, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
cnapagoda: For new file, year should be 2015

PR #169, opened 2015-05-20
adding custom sequences to resources folder

malinthaprasan:


modules/distribution/resources/customsequences/fault/json_fault.xml
```diff
@@ -0,0 +1,41 @@
...
+    <filter xpath="$ctx:CUSTOM_HTTP_SC">
+        <then>
+            <property name="HTTP_SC" expression="$ctx:CUSTOM_HTTP_SC" scope="axis2"/>
+        </then>
+        <else>
+            <property name="HTTP_SC" value="500" scope="axis2"/>
+        </else>
+    </filter>
+    <class name="org.wso2.carbon.apimgt.usage.publisher.APIMgtFaultHandler"/>
+    <property name="RESPONSE" value="true"/>
```
chanakaudaya: you can use <respond/> mediator in place of following 4 mediators
<property name="RESPONSE" value="true"/>
<header name="To" action="remove"/>
<send/>
<drop/>

PR #307, opened 2015-08-03
remove mediation/synapse from store/pub

dhanuka84:


harsha89:
Closed as no longer valid

PR #608, opened 2016-01-14
Fix PostgreSQL script

uvindra:


chamilaadhi:
This branch is already released. Can you do a PR to master branch please?

uvindra:
This is a migration client fix so we have to commit to this branch

PR #618, opened 2016-01-18
Update usersignup BPEL for APIMANAGER-2994 fix

praminda:
correlation and exit event listener was added in order to implement
task cleanup process for user sign up.


hasithaa:
Since you are updating the BPEL file inside the zip file, your changes will not recorded in git version history. But having version history of the BPEL file is very useful for trouble shooting. 

IMO correct way is convert these artifacts to BPEL maven projects and build them at build time. Do the same for HumanTask as well.

praminda:
Added [APIMANAGER-4602](https://wso2.org/jira/browse/APIMANAGER-4602) to fix at a later time.

PR #1132, opened 2016-10-10
Adding automated test for SDK generation feature.

chamindias:


sajinidesilva:
I think we can squash commits before merging in PRs like this.

PR #1244, opened 2016-12-08
added xacml mediation

sineth-neranjana:

PR #1246, opened 2016-12-08
Added XACML mediation with correct XACML mediation feature

sineth-neranjana:


chamindias:
This is to make some corrections related to : https://github.com/wso2/product-apim/pull/1244 by Sineth.

PR #1294, opened 2017-01-11
Support templates in api.json file when importing

jmalvarezf:
Hi,

Again as part of the project here in "El Corte Ingles" in Spain, we have developed an extension in the import-export api war app to be able to apply velocity templates in the api.json file, when importing an API. That way, we would be able to support what we would call "api as code", and we would be able to define apis programatically, and define the backend URL based on configuration systems like puppet, chef, or similar. 

We have done a few modifications:

- We have modified the import process to get all the HTTP headers that start with "X-WSO2" prefix, and get them and their values in a map that will serve as configuration values to query the configuration system. 
- We have provided an interface to be able to implement different ways to get configuration keys from configuration systems. We have developed one of our own, but as the configuration system is ad-hoc, it doesn't add much.
- We would have to deploy the jar files from the interface and the implementation in the components/lib directory, and the war in the same way it is actually done.

If you have any question, please don't hesitate to ask.

Thank you,

Jose Maria.


harsha89:
This pull request is invalid. Hence closing it.

jmalvarezf:
Hello @harsha89,

Why is it invalid? What's the problem? It is trying to solve a problem that WSO2 has when trying to import the same API into different environments, or even, trying to treat the API specification as code and introduce it into continuous deployment. It creates an interface to be able to integrate, for example, puppet, to recover configuration details for the API import configuration.

Please, share your comments.

Thank you!

Jose Maria.

harsha89:
@jmalvarezf 

We have moved the master to newest C5 developments. Hence the API Import and Export is not handler part of our rest API in the https://github.com/wso2/carbon-apimgt.

PR #1302, opened 2017-01-19
Hide unnecessary identity menu items from management console

praminda:


praminda:
https://wso2.org/jira/browse/APIMANAGER-5589

PR #1427, opened 2017-06-23
Included Apache ActiveMQ requirement

lkamal:
ActiveMQ is required for API Manager 3.0.0-m3.


sajhak:
Thanks @lkamal for the PR, unfortunately this seems to have gone unnoticed, hence not merged. However the ActiveMQ requirement has been added in later README s, hence this PR is outdated, and therefore I'm going to close this. But, please continue to try out and evaluate API Manager releases and we would like to see more contributions in the future :)

PR #1503, opened 2017-08-22
Merging changes from master

shaniR:


analyzer/pom.xml
```diff
@@ -1078,7 +1078,7 @@
     </build>
 
      <properties>
-        <carbon.apim.version>7.0.37-SNAPSHOT</carbon.apim.version>
+        <carbon.apim.version>7.0.39</carbon.apim.version>
```
harsha89: this should be 7.0.40-SNAPSHOT

shaniR: should be fixed in master right?

anuruddhal: shouldn't this be 7.0.40-SNAPSHOT?

harsha89:
@anuruddhal we need to fix this in master

PR #1698, opened 2017-11-16
Added sample eight

prasa7:
This pull request consists of code related to sample scenario eight.

### DIFF
--- a/sample-scenarios/sample8/src/main/java/CreateSampleEightRawData.java
+++ b/sample-scenarios/sample8/src/main/java/CreateSampleEightRawData.java
@@ -33,7 +33,7 @@ public class CreateSampleEightRawData {
     private static final String port = "9443";
 
     /**
-     * This main method will be called when running sample one.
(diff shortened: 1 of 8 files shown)


sample-scenarios/sample8/src/main/java/CreateSampleEightRawData.java
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * This class is used to populate sample data to represent the sample eight business scenario.
+ */
+public class CreateSampleEightRawData {
+
+    private static final String hostname = "localhost";
+    private static final String port = "9443";
+
+    /**
+     * This main method will be called when running sample one.
```
chamilaadhi: should be sample 8 right? :)

prasa7: Thanks for pointing out. FIxed in e29065f4d11f91950e1f30d35d5cfdc1acdf6655

PR #2725, opened 2018-02-12
Fix https://github.com/wso2/carbon-transports/issues/433

chamindias:


malinthaprasan:
As the PR contains hardcoded JKS configs, for someone who update default keystore paswords etc it is likely to miss this new config. Shall we update [1] with that information?
[1] https://docs.wso2.com/display/AM210/Configuring+Keystores+in+WSO2+API+Manager

PR #3359, opened 2018-06-08
Validate files in zip archives

msm1992:
## Purpose
This PR adds archive file content validation for API import feature.


modules/api-import-export/src/main/java/org/wso2/carbon/apimgt/importexport/utils/APIImportUtil.java
```diff
@@ -155,6 +155,12 @@ public static String extractArchive(File sourceFile, String destination) throws
 
                 File destinationFile = new File(destination, currentEntry);
                 File destinationParent = destinationFile.getParentFile();
+                String canonicalizedDestinationFilePath = destinationFile.getCanonicalPath();
+                if (!canonicalizedDestinationFilePath.startsWith(new File(destination).getCanonicalPath())) {
+                    String errorMessage = "Entry is outside of the target dir: " + currentEntry;
```
bhathiya: Should we mention that the zip is invalid and the reason is xxxx?

PR #3502, opened 2018-07-02
Fixes for integration test issue on product-apim 2.x

shavantha:
The three test cases below only created the API/Application and the Subscription and when  the test ran more than once the test had failed with duplicate error. With this changes on the destroy method we remove the subscription/application and the api
## Approach
Publisher :APIResourceWithTemplateTestCase.java  
          
WebSocket :APIMANAGER5869WSGayewatURLTestCase.java
           WebSocketAPITestCase.java  

Header :CORSHeadersTestCase.java 

The below test case had an assertion that was always successful but the assertion on line 116 was looking for a failure and was failing 
APIM614AddDocumentationToAnAPIWithDocTypeSampleAndSDKThroughPublisherRestAPITestCase


shavantha:
closing till the issues are fixed

PR #3548, opened 2018-07-19
fix file path and resource path separator for all the env including windows

chaminda:
## Purpose
This will fix the integration test failures occurred on Windows platform.  

## Approach
corrected file url prefix in WebAppAdminClient#uploadWarFile as "File:///"
previous value "File://" doesn't work on Windows environments

corrected resource path separater as "/" in DefaultEndpointTestCase#testAPIInvocationAfterAddingDynamicEndpoint
previous value was the File.separator which led the url to be malformed in Windows.

## Test environment
Tested on Centos, Ubuntu, Windows


chaminda:
@rmsamitha, yes we have used the same in different other locations as well. Ex:- APIMIntegrationBaseTest

PR #3567, opened 2018-07-25
Fixing File Path Issue With Escape Character

maneeshaheshan:
Previous value "[\\\\/]" doesn't work on Windows environments.

> Corrected relativeFilePath in ESBJAVA3380TestCase#setEnvironment as "[\\\\]".
Previous value "[\\\\/]" doesn't work on Windows environments.


maneeshaheshan:
Changes done to remove the path separator replacement from the tests and tested with Windows, Centos and Ubuntu.

PR #3572, opened 2018-07-30
Integration tests

tharindu1st:
## Purpose

This adds the integration tests for the APIM 3.x


modules/integration-tests/src/test/java/org/wso2/carbon/apimgt/rest/integration/tests/store/APISubscriptionTestCaseIT.java
```diff
@@ -0,0 +1,179 @@
...
+        Subscription subscriptionNegative = new Subscription().apiIdentifier(apiId).applicationId(applicationId)
+                .policy("Bronze1");
+        try {
+            subscriptionIndividualApi.subscriptionsPost(subscriptionNegative);
+            Assert.fail();
+        } catch (Exception e) {
+            if (e.getCause() instanceof RestAPIException) {
+                RestAPIException restAPIException = (RestAPIException) e.getCause();
+                Error error = restAPIException.getError();
+                Assert.assertEquals(error.getMessage(), "Policy Not found");
```
uvindra: This level of detail is good but if we do any changes to the formatting of the exception message it will lead to test failures. Its bets if we can use a constant to define the error message string and reuse the same in the code and test

modules/integration-tests/src/test/java/org/wso2/carbon/apimgt/rest/integration/tests/store/APISubscriptionTestCaseIT.java
```diff
@@ -0,0 +1,179 @@
...
+                Assert.assertEquals(moreInfo.get("APPLICATION_ID"), applicationId);
+                Assert.assertEquals(moreInfo.get("TIER"), "Bronze1");
+            } else {
+                Assert.fail();
+            }
+        }
+    }
+
+    @AfterClass
+    public void destroy() {
```
uvindra: Any reason why we are doing this explicitly since we are anyway dropping the tables after each test run?

modules/integration-tests/pom.xml
```diff
@@ -94,6 +94,25 @@
                 <artifactId>maven-javadoc-plugin</artifactId>
                 <version>2.10.4</version>
```
malinthaprasan: shall we move this to the parent pom

modules/integration-tests/pom.xml
```diff
@@ -94,6 +94,25 @@
...
             </plugin>
+            <plugin>
+                <groupId>io.swagger</groupId>
+                <artifactId>swagger-codegen-maven-plugin</artifactId>
+                <version>2.2.3</version>
```
malinthaprasan: This one too.

PR #3704, opened 2018-09-10
Fix tests cases and update master-ds.xml

bhathiya:
## Purpose

- Take all Thread.sleeps, which were added to make sure 2 lifecycle states have 2 different timestamps, to a single place
- Set default autocommit true in master ds


Mariangela:
@bhathiya & @chamilaadhi , please mark if this has a doc impact or not by selecting the appropriate label (Docs/Has Impact OR Docs/No Impact).

bhathiya:
This has a doc impact and I already asked @shaniR to make the changes.

PR #3934, opened 2018-11-05
Fix POM files, and Improve the APINewVersionTestCase

kavithasub:
## Purpose
- Fix POM files
- Change new name as APINewVersionTestCase for initial test case
- Improve the Fix test case with more methods


product-scenarios/1-api-updates-using-new-versions/pom.xml
```diff
@@ -64,11 +59,6 @@
                 <jdk>1.4</jdk>
             </activation>
             <dependencies>
-                <dependency>
```
Krishanx92: We can remove all under <profiles> section. Not only dependencies.

PR #3951, opened 2018-11-08
Add new test script RESTApiCreationWithOptionalFiledsTestCase.java and override constructor APIRequest

aaujayasena:
Add new test script for test case ID:   1.1.1.5
Test case description: An API creator should be able to create a REST API from scratch by
providing all the mandatory and optional fields.

### DIFF
--- a/product-scenarios/1-manage-public-partner-private-apis/1.1-expose-service-as-rest-api-and-apply-qos/1.1-common/src/main/java/org/wso2/am/scenario/test/common/APIRequest.java
+++ b/product-scenarios/1-manage-public-partner-private-apis/1.1-expose-service-as-rest-api-and-apply-qos/1.1-common/src/main/java/org/wso2/am/scenario/test/common/APIRequest.java
@@ -40,9 +40,7 @@ public class APIRequest extends AbstractRequest {
     }
 
     @Override
-    public void setAction() {
-        this.setAction("design");
-    }
+    public void setAction() {}
 
     @Override
     public void init() {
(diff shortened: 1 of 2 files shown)


product-scenarios/1-manage-public-partner-private-apis/1.1-expose-service-as-rest-api-and-apply-qos/1.1.1-create-rest-api-from-scratch/src/test/java/org/wso2/am/scenario/tests/rest/api/creation/RESTApiCreationWithOptionalFiledsTestCase.java
```diff
@@ -0,0 +1,75 @@
...
+import org.testng.annotations.Test;
+import org.wso2.am.integration.test.utils.APIManagerIntegrationTestException;
+import org.wso2.am.scenario.test.common.APIPublisherRestClient;
+import org.wso2.am.scenario.test.common.APIRequest;
+import org.wso2.am.scenario.test.common.ScenarioTestBase;
+import org.wso2.carbon.automation.test.utils.http.client.HttpResponse;
+
+import java.util.Properties;
+
+public class RESTApiCreationWithOptionalFiledsTestCase extends ScenarioTestBase {
```
fazlan-nazeem: I don' think we need a separate class for this test case. Shall we move this to the already existing class?

product-scenarios/1-manage-public-partner-private-apis/1.1-expose-service-as-rest-api-and-apply-qos/1.1-common/src/main/java/org/wso2/am/scenario/test/common/APIRequest.java
```diff
@@ -51,11 +53,26 @@ public void init() {
         this.addParameter("visibility", this.visibility);
         this.addParameter("version", this.version);
         this.addParameter("swagger", this.swagger);
+        this.addParameter("tags", this.tags);
```
fazlan-nazeem: We are not passing tags, description as arguments in this constructor. So we shouldn't be initializing them here.

PR #3952, opened 2018-11-08
Add cross tenant API export support for super admin for migration 

rmsamitha:
## Purpose
Adding cross tenant API export support into the api-import-export REST API. This is for the objective of APIM 2.6.0 to 3.0.0 migration. 

## Approach
This support is enabled under the conditions that the caller is a super admin user and the server is started with migrationMode=true system property. And this is implemented by starting tenantFlow if the above conditions are satisfied.

### DIFF
--- a/modules/api-import-export/src/main/java/org/wso2/carbon/apimgt/importexport/utils/APIExportUtil.java
+++ b/modules/api-import-export/src/main/java/org/wso2/carbon/apimgt/importexport/utils/APIExportUtil.java
@@ -19,6 +19,7 @@
 package org.wso2.carbon.apimgt.importexport.utils;
 
 import org.wso2.carbon.apimgt.impl.definitions.APIDefinitionFromOpenAPISpec;
+import org.wso2.carbon.apimgt.impl.internal.ServiceReferenceHolder;
 import org.wso2.carbon.apimgt.importexport.APIExportException;
 
 import com.google.gson.Gson;
@@ -51,6 +52,9 @@ import org.wso2.carbon.registry.api.RegistryException;
 import org.wso2.carbon.registry.api.Resource;
 import org.wso2.carbon.registry.core.RegistryConstants;
 import org.wso2.carbon.registry.core.service.RegistryService;
+import org.wso2.carbon.user.api.UserStoreException;
+import org.wso2.carbon.utils.multitenancy.MultitenantConstants;
+import org.wso2.carbon.utils.multitenancy.MultitenantUtils;
 
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
@@ -194,7 +198,7 @@ public class APIExportUtil {
         //export documents
         List<Documentation> docList;
         try {
-            docList = provider.getAllDocumentation(apiID);
+            docList = provider.getAllDocumentation(apiID, userName);
         } catch (APIManagementException e) {
             String errorMessage = "Unable to retrieve API Documentation";
             log.error(errorMessage, e);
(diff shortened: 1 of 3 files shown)


modules/api-import-export/src/main/java/org/wso2/carbon/apimgt/importexport/utils/APIExportUtil.java
```diff
@@ -812,4 +821,73 @@ private static void writeFile(String path, String content) throws APIExportExcep
...
+        if (resourceRquesterDomain.equals(apiDomain)) {
+            return false;
+        }
+        String superAdminRole;
+        try {
+            superAdminRole = ServiceReferenceHolder.getInstance().getRealmService().
+                    getTenantUserRealm(MultitenantConstants.SUPER_TENANT_ID).getRealmConfiguration().getAdminRoleName();
+        } catch (UserStoreException e) {
+            String errorMsg = "Error in getting super admin role name";
+            throw new APIExportException(errorMsg, e);
```
npamudika: Don't we need to log the exception here?

modules/api-import-export/src/main/java/org/wso2/carbon/apimgt/importexport/utils/APIExportUtil.java
```diff
@@ -812,4 +821,73 @@ private static void writeFile(String path, String content) throws APIExportExcep
...
+        }
+
+        //check whether logged in user is a super tenant user
+        String superTenantDomain;
+        try {
+            superTenantDomain = ServiceReferenceHolder.getInstance().getRealmService().getTenantManager().getSuperTenantDomain();
+        } catch (UserStoreException e) {
+            String errorMsg = "Error in getting the super tenant domain";
+            throw new APIExportException(errorMsg, e);
```
npamudika: It is good if we log the error and then throw the exception.

modules/api-import-export/src/main/java/org/wso2/carbon/apimgt/importexport/utils/APIExportUtil.java
```diff
@@ -812,4 +821,73 @@ private static void writeFile(String path, String content) throws APIExportExcep
...
+        if(!isSuperTenantUser) {
+            return true;
+        }
+        //check whether the user has super tenant admin role
+        boolean isSuperAdminRoleNameExist;
+        try {
+            isSuperAdminRoleNameExist = isUserInRole(username, superAdminRole);
+        } catch (UserStoreException e) {
+            String errorMsg = "Error in checking whether the user" + username + " has admin role";
+            throw new APIExportException(errorMsg, e);
```
npamudika: Same here.

modules/api-import-export/src/main/java/org/wso2/carbon/apimgt/importexport/utils/APIExportUtil.java
```diff
@@ -812,4 +821,73 @@ private static void writeFile(String path, String content) throws APIExportExcep
...
+        } catch (APIManagementException e) {
+            String errorMsg = "Error in checking whether the user" + username + " has admin role";
+            throw new APIExportException(errorMsg, e);
```
npamudika: Same here.

modules/api-import-export/src/main/java/org/wso2/carbon/apimgt/importexport/utils/APIExportUtil.java
rmsamitha: Logging and then throwing will cause logging the same error message multiple times. Error is 'catched' here to pass information about the particular error occurred in the try block, to the upper level, where it is logged.

modules/api-import-export/src/main/java/org/wso2/carbon/apimgt/importexport/utils/APIExportUtil.java
rmsamitha: Logging and then throwing will cause logging the same error message multiple times. Error is 'catched' here to pass information about the particular error occurred in the try block, to the upper level, where it is logged.

modules/api-import-export/src/main/java/org/wso2/carbon/apimgt/importexport/utils/APIExportUtil.java
rmsamitha: Same reason mentioned above

modules/api-import-export/src/main/java/org/wso2/carbon/apimgt/importexport/utils/APIExportUtil.java
rmsamitha: Same reason mentioned above

PR #3963, opened 2018-11-13
Product scenarios

aaujayasena:
Add new scenario test method 1.1.1.5- testRESTAPICreationWithOptionalValues.


product-scenarios/1-manage-public-partner-private-apis/1.1-expose-service-as-rest-api-and-apply-qos/common-utils/src/main/java/org/wso2/am/scenario/test/common/APIRequest.java
```diff
@@ -85,6 +139,64 @@ public void init() {
         if (endpoint != null) {
             this.addParameter("endpoint_config", endpoint.toString());
         }
+        if(this.tierAvailability != null){
```
Krishanx92: Formatting issues. please fix other places as well.

Issue #3743, opened 2018-09-16
Non admin users cannot search for an API using custom properties.


dushaniw:
**Description:**
Non admin users cannot search APIs using custom API properties <property_name>:<property_value> in Publisher. However search results are returned successfully for admin users.

To reproduce,

1. Create a user with Internal/creator role
2. Login to publisher with credentials of this user and create an API with a custom API property and save the API.
3. Search the API using a similar query as, "<property_name>:<property_value>"
   Eg: "environment:preprod"
4. "Info! No matching APIs found." is returned with 0 search results.

Try the step 3 after log in as an admin user and the created API will be visible in search results.

**Suggested Labels:**
Bug 

**Affected Product Version:**
2.5.0
2.6.0

**OS, DB, other environment details and versions:**    
Ubuntu 16.04
Default H2 DB/ Microsoft SQL Server 2017

PR #3988, opened 2018-11-23
Changed the indexer to fix the error in API property search

HiranyaKavishani:
## Purpose
This fixes for the API custom property search issue for non-admin users.

## Approach
Configure the CustomAPIIndexer instead of the RXTIndexer, in where the API properties were added as attributes while indexing, so it can searches properties from the registry by solr attribute search

https://github.com/wso2/product-apim/issues/3743


bhathiya:
Fixed by https://github.com/wso2/carbon-apimgt/pull/5866/files

PR #4021, opened 2018-11-29
Product scenarios

fazlan-nazeem:
## Purpose

Adds scenario 9 to root pom
Modify readme to use relative paths instead of absolute URL.
Add pom to child nodes


product-scenarios/pom.xml
```diff
@@ -39,6 +39,7 @@
         <module>5-managing-api-lifecycle</module>
         <module>8-analyze-usage-of-apis-for-business-insights</module>
         <module>14-receive-alerts-on-different-events-for-decision-making</module>
+        <module>9-developer-community-building</module>
```
bhathiya: shall we maintain the order?

fazlan-nazeem: @bhathiya This is fixed

PR #4059, opened 2018-12-06
Fix Publishing new Versions of API's with "Deprecate Old Versions" th…

nimanthag:


nimanthag:
### Developer testing record
Followed the steps in issue and. New version added without any error.
screen shot 2018-12-06 at 5 19 37 pm

PR #4061, opened 2018-12-06
2.x fix jwt x5t

### LINKED ISSUE #5535: Incorrect paths in APIM 2.6.0 documentation when configuring app creation workflow
Multiple path errors in following doc - https://docs.wso2.com/display/AM260/Adding+an+Application+Creation+Workflow#UsingEI

Under "Configuring the Business Process Server"
- Step 4 - Correct paths should be <EI_HOME>/**wso2**/business-processes/epr/...
- Step 5 - It should explicitly mention to start business process server using <EI_HOME>/bin/business-process.sh. General instruction "Start the EI server and sign in..." is not correct.

gusto2:
## Purpose

Fixing wso2 #5884, mitigating incorrect fix #5535, updated testcase to check for url-encoded x5t jwt header. The test case is aligned with pull https://github.com/wso2/carbon-apimgt/pull/5886

## Goals

Generate validable JWT backend token

## Approach

Encodings of all JWT elements were done using basic Base64 (java.util.Base64) with URL encoding and without wrapping. That made the generated token and signature verfiable by 3rd party tools (usable by backend clients)

## Security checks

"x5t" jwt header is considered obsolete (using sha-1), we should use "x5t#256" or make it configurable

## Test environment

oracle jdk 1.8_181, Amazon Linux2 (CentOS)

### DIFF
--- a/modules/integration/tests-common/integration-test-utils/src/main/java/org/wso2/am/integration/test/utils/generic/APIMTestCaseUtils.java
+++ b/modules/integration/tests-common/integration-test-utils/src/main/java/org/wso2/am/integration/test/utils/generic/APIMTestCaseUtils.java
@@ -1566,12 +1566,15 @@ public class APIMTestCaseUtils {
 
     public static boolean isJwtSignatureValid(String jwtAssertion, byte[] jwtSignature, String jsonHeader) throws UnsupportedEncodingException {
         KeyStore keyStore = null;
-        String thumbPrint = null;
+        byte[] thumbPrint = null;
         String signatureAlgorithm = null;
         JSONObject jsonHeaderObject = null;
         try {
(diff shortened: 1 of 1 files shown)


bhathiya:
@gusto2 First of all thank you very much for the PR and my apologies for the delay to respond. 

We went through the requirement again and tested with JWT.io and Nimbus library too. It worked when we change the base64 to base64URL. However, we believe that certificate thumbprint should be a hex value[1]. We tested that too and it works fine with the above 2. 

Therefore, unfortunately, we're unable to merge your PR as it is. However, we fixed the base64>base64url a few days back. 

Anyways, thank you again for the PR and please keep sending them. :) 

[1] https://support.konghq.com/s/case/5001K00000ltyrA

PR #4138, opened 2019-01-07
Modifying waiting logic

fazlan-nazeem:
## Purpose

Does an API get call prior to the subscription call. if API is not available in get call, then the subscription will skip and th test will fail.

### DIFF
--- a/product-scenarios/scenarios-common/src/main/java/org/wso2/am/scenario/test/common/ScenarioTestBase.java
+++ b/product-scenarios/scenarios-common/src/main/java/org/wso2/am/scenario/test/common/ScenarioTestBase.java
@@ -275,31 +275,35 @@ public class ScenarioTestBase {
         }
     }
 
-    public void isAPIVisibleInStoreForAnonymousUser(String apiName, String tenantDomain)
+    public void isAPIVisibleInStore(String apiName, APIStoreRestClient apiStoreRestClient)
             throws APIManagerIntegrationTestException {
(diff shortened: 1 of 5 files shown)


product-scenarios/scenarios-common/src/main/java/org/wso2/am/scenario/test/common/ScenarioTestBase.java
```diff
@@ -275,32 +275,36 @@ public void deleteRole(String role, String adminUsername, String adminPassword)
         }
     }
 
-    public void isAPIVisibleInStoreForAnonymousUser(String apiName, String tenantDomain)
+    public void isAPIVisibleInStore(String apiName, APIStoreRestClient apiStoreRestClient)
             throws APIManagerIntegrationTestException {
-        APIStoreRestClient apiStoreClient = new APIStoreRestClient(storeURL);
-
-        long waitTime = System.currentTimeMillis() + 5 * 1000;
+        long waitTime = System.currentTimeMillis() + ScenarioTestConstants.WAIT_TIME;
```
malinthaprasan: shall we use a bit specific name for this: suggestion: API_APPEAR_IN_STORE_TIMEOUT_MILLIS

fazlan-nazeem: Ack. will send the change with another PR

PR #4283, opened 2019-02-01
Add 2.1.1.7 and 2.1.1.8 testcases

HiranyaKavishani:
## Purpose
Add scenario test case for check an API publisher would not be able to publish an API when the required permission or role removed

### DIFF
--- a/product-scenarios/scenarios-common/src/main/java/org/wso2/am/scenario/test/common/ScenarioTestBase.java
+++ b/product-scenarios/scenarios-common/src/main/java/org/wso2/am/scenario/test/common/ScenarioTestBase.java
@@ -246,7 +246,8 @@ public class ScenarioTestBase {
         }
     }
 
-    public void createRole(String adminUsername, String adminPassword, String role, String[] permisionArray) throws APIManagementException {
+    public void createRole(String adminUsername, String adminPassword, String role,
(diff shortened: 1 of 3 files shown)


product-scenarios/scenarios-common/src/main/java/org/wso2/am/scenario/test/common/ScenarioTestBase.java
```diff
@@ -261,6 +262,23 @@ public void createRole(String adminUsername, String adminPassword, String role,
...
     }
 
+    public void updateRole(String adminUsername, String adminPassword, String role, String[] userList,
+                           String[] permissionArray) throws APIManagementException {
+
+        UserManagementClient userManagementClient = null;
+        try {
+            userManagementClient = getRemoteUserManagerClient(adminUsername, adminPassword);
+            userManagementClient.deleteRole(role);
+            userManagementClient.addRole(role,
```
isharac: Is the permissionArray contains (existingPermissions + newPermissions - removingPermissions)?

HiranyaKavishani: We remove the existing permissions from the role which assigned for a user and replaces it with the new permission

PR #4286, opened 2019-02-01
Add a common method to develop/publish sample APIs for scenario tests

rmsamitha:
## Purpose
The changes are done to add a common method to develop/publish APIs for scenario tests, so it can be reused for other tests.
This method does the designing, implementing and managing (and publish if required) an API via the approach that the Publisher UI is performing, by calling relevant required Jaggery APIs.


modules/integration/tests-common/integration-test-utils/src/main/java/org/wso2/am/integration/test/utils/bean/APIManageBean.java
```diff
@@ -0,0 +1,284 @@
...
+package org.wso2.am.integration.test.utils.bean;
+
+public class APIManageBean extends AbstractRequest {
+    private String name;
+    private String type = "";
+    private String version = "";
+    private String provider = "";
+    private String default_version_checked = "";
+    private String transport_https = "";
+    private String transport_http = "";
```
praminda: shall we use camel case for all variables in here

modules/integration/tests-common/integration-test-utils/src/main/java/org/wso2/am/integration/test/utils/bean/APIManageBean.java
```diff
@@ -0,0 +1,284 @@
...
+
+    public String getProvider() {
+        return provider;
+    }
+
+    public void setProvider(String provider) {
+        this.provider = provider;
+    }
+
+    public String getDefault_version_checked() {
```
praminda: Change getters/setters to camel case

product-scenarios/pom.xml
```diff
@@ -142,7 +142,7 @@
             <dependency>
                 <groupId>org.wso2.am</groupId>
                 <artifactId>org.wso2.am.integration.common.test.utils</artifactId>
-                <version>${apimserver.version}</version>
+                <version>2.6.1-SNAPSHOT</version>
```
praminda: Any reason to change this here?

product-scenarios/scenarios-common/src/main/java/org/wso2/am/scenario/test/common/ScenarioTestBase.java
```diff
@@ -441,4 +446,67 @@ public boolean isTagsResponseContainsTag(JSONObject tagsResponse, String tagName
...
+     *
+     * @param apiPublisher APIPublisherRestClient
+     * @param swaggerFileRelativePath Swagger file relative path to create API
+     * @param apiDeveloperUsername Developer name
+     * @param backendEndPoint Backendpoint url
+     * @param publish Whether to publish APi or not
+     * @param storeVisibility Store visibility level of the API
+     * @throws Exception
+     */
+    public void developSampleAPI(APIPublisherRestClient apiPublisher, String swaggerFileRelativePath,
```
praminda: I'm wondering why this method should be in here instead of APIPublisherRestClient?

modules/integration/tests-common/integration-test-utils/src/main/java/org/wso2/am/integration/test/utils/bean/APIManageBean.java
rmsamitha: Form params are sent exactly as above (with underscores), from UI. We can still change these as they carry only the value, but keeping consistent everywhere might make more clear. WDYT ?

product-scenarios/scenarios-common/src/main/java/org/wso2/am/scenario/test/common/ScenarioTestBase.java
rmsamitha: We have defined APIPublisherRestClient to add methods to directly call particular Publisher Jaggery REST APIs. But this developSampleAPI(..) method is different right? It makes multiple REST calls as well do other operations to achieve the objective of developing a sample API. Methods currently existing within ScenarioTestBase are doing such that various operations, and they are using multiple Rest clients.

praminda: IMO since it is a Rest client functionality more than a utility it will be better if we can have it in RestClient side

rmsamitha: Changed with cd65577

modules/integration/tests-common/integration-test-utils/src/main/java/org/wso2/am/integration/test/utils/bean/APIManageBean.java
rmsamitha: Fixed with cd655777f335d12a56f15e377637c8ebf0f08e7a

PR #4326, opened 2019-02-11
Test case changes after removing the provider from request params

tmkasun:


chamilaadhi:
lets upgrade the carbon.apimgt version as well since the changes related to this is comming from the those components

PR #4413, opened 2019-03-06
Product scenarios dev

mushthaq33:
## Purpose
Fixing Test failures


product-scenarios/scenarios-common/src/main/java/org/wso2/am/scenario/test/common/APIPublisherRestClient.java
```diff
@@ -977,7 +978,6 @@ public void developSampleAPI(JSONObject swaggerJson,
                 HttpResponse apiPublishResponse = changeAPILifeCycleStatus(updateLifeCycle);
                 scenarioTestBase.verifyResponse(apiPublishResponse);
             }
-            Thread.sleep(15000);
```
praminda: Shall we keep this here till we implement a proper solution

mushthaq33: +1. Fixed with ec0bc7d80f8664d001910d8a439e58506a7059e1

PR #4480, opened 2019-03-25
Improvements for build failures

praminda:
## Purpose
Add new improvements to prevent recent build failures

### Goals
* Add API monitoring webapp to wait for API deployment
* Fix possible causes for test failures


product-scenarios/scenarios-common/src/main/java/org/wso2/am/scenario/test/common/ScenarioTestBase.java
```diff
@@ -587,4 +624,89 @@ public boolean isTagsResponseContainsTag(JSONObject tagsResponse, String tagName
...
+                log.info("Data: " + response.getData());
+                if (response.getData().contains(expectedResponse)) {
+                    log.info("API :" + apiName + " with version: " + apiVersion +
+                            " with expected response " + expectedResponse + " found");
+                    break;
+                } else {
+                    try {
+                        Thread.sleep(500);
+                    } catch (InterruptedException ignored) {
+
```
dinusha92: can we handle this exception in a proper way?

praminda: @dinusha92 we need to ignore this exception. What we are doing here is trying to wait for API deployment in a loop. So we are pausing the loop for 500ms after checking the deployment status onece. Failure to sleep is ignored and no need to log/handle it. Anyway I'll add a comment in the code

PR #4684, opened 2019-05-08
Removed optional scope validator from identity.xml

HiranyaKavishani:
## Purpose
Removed Optional Scope Validations (<ScopeValidators>) from identity.xml because by default APIM use  <OAuthScopeValidator> to validate scopes

### DIFF
--- a/modules/distribution/product/identity_config_change.xml
+++ b/modules/distribution/product/identity_config_change.xml
@@ -12,6 +12,11 @@
       <after>//s:Server/s:OAuth/s:OAuthCallbackHandlers</after>
       <value><![CDATA[<OAuthScopeValidator class="org.wso2.carbon.identity.oauth2.validators.JDBCScopeValidator"/>]]></value>
    </add>
+
+    <remove>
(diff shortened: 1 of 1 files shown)


modules/distribution/product/identity_config_change.xml
```diff
@@ -12,6 +12,11 @@
       <after>//s:Server/s:OAuth/s:OAuthCallbackHandlers</after>
       <value><![CDATA[<OAuthScopeValidator class="org.wso2.carbon.identity.oauth2.validators.JDBCScopeValidator"/>]]></value>
    </add>
+
+    <remove>
+        <name>//s:Server/s:OAuth/s:ScopeValidators</name>
+    </remove>
```
TanyaM: formatting issue with indentation

HiranyaKavishani: The added part doesn't have any formatting issue. Remains L2-L4, L6-L11, L13-L17, L23-L30 lines have the formatting issue

PR #4927, opened 2019-05-28
Disable carbon metrics

fazlan-nazeem:


fazlan-nazeem:
closing since jdbc log metrics has already been disabled.

PR #5155, opened 2019-07-04
Add db scripts for migration from 2.6.0 to 3.0.0

vithu30:
This PR adds AM_DB migration scripts and data migration scripts for changing the default static configuration value as false, for migrating from 2.6.0 to 3.0.0


modules/distribution/resources/migration-2.6.0_to_3.0.0/db2.sql
```diff
@@ -0,0 +1,46 @@
...
+/
+CREATE TRIGGER AM_SYSTEM_APPS_TRIGGER NO CASCADE BEFORE INSERT ON AM_SYSTEM_APPS
+REFERENCING NEW AS NEW FOR EACH ROW MODE DB2SQL
+  BEGIN ATOMIC
+    SET (NEW.ID)
+    = (NEXTVAL FOR AM_SYSTEM_APPS_SEQUENCE);
+  END
+/
+
+CREATE TABLE AM_API_CLIENT_CERTIFICATE (
```
harsha89: Shouldn't we use create table if not exist?

modules/distribution/resources/migration-2.6.0_to_3.0.0/h2.sql
```diff
@@ -0,0 +1,37 @@
+CREATE TABLE IF NOT EXISTS AM_SYSTEM_APPS (
```
harsha89: should these manually executed?

modules/distribution/resources/migration-2.6.0_to_3.0.0/versioning-disable-scripts/h2.sql
```diff
@@ -0,0 +1,40 @@
+-- Update the REG_PATH_ID column mapped with the REG_RESOURCE table --
+
```
harsha89: We will need to add a section to migration document? Do we have a draft version of it otherwise we might miss it.  Again customers in existing version may need this if they decided to migrate. We need to carefully handle this and properly documented.

modules/distribution/resources/migration-2.6.0_to_3.0.0/db2.sql
vithu30: DB2 does not support IF NOT EXISTS statement. This check is not there on other DB2 scrips as well..

modules/distribution/resources/migration-2.6.0_to_3.0.0/h2.sql
vithu30: yes. we need to execute this manually

modules/distribution/resources/migration-2.6.0_to_3.0.0/versioning-disable-scripts/h2.sql
vithu30: [1] is the draft version which does not contains complete steps as we need to finalize on IS component migration. Further, I have created [2], [3] to be documented which are identified during testing.
[1] https://docs.google.com/document/d/1QVDL3sUFxQxMNNqPF-WHe7FhM6SR0gvVLzdECQPYqz8/edit?usp=sharing
[2] https://github.com/wso2/product-apim/issues/5183
[3] https://github.com/wso2/product-apim/issues/5182

PR #5166, opened 2019-07-07
Added support for YAML

kasvith:
In following PR
- Added YAML exporting support for APIs via a query parameter called `format`
- Default exporting format is set to `json` (can be changed and must go parallel with tooling)
- Added support for importing APIs in YAML or JSON format

> Note: When importing always YAML files treated first and then as a fallback JSON. This is because of the latest tooling pre-process API and store them in YAML format. In future, we would simply remove JSON format and default to YAML


modules/api-import-export/src/main/java/org/wso2/carbon/apimgt/importexport/utils/APIImportUtil.java
```diff
@@ -190,6 +193,102 @@ public static String extractArchive(File sourceFile, String destination) throws
...
+                Node node = states.item(i);
+                Node id = node.getAttributes().getNamedItem("id");
+                if (id != null && !id.getNodeValue().isEmpty()) {
+                    LifeCycleTransition lifeCycleTransition = new LifeCycleTransition();
+                    NodeList transitions = node.getChildNodes();
+                    int nTransitions = transitions.getLength();
+                    for (int j = 0; j < nTransitions; j++) {
+                        Node transition = transitions.item(j);
+                        // Add transitions
+                        if (transition.getNodeName().equals("transition")) {
```
harsha89: contant.equalts(variable) pattern

modules/api-import-export/src/main/java/org/wso2/carbon/apimgt/importexport/utils/APIImportUtil.java
```diff
@@ -343,6 +468,11 @@ public static void importAPI(String pathToArchive, String currentUser, boolean i
                 provider.updateAPI(importedApi);
             }
 
+            // Change API lifecycle if state transition is required
+            if (lifecycleAction != null && !lifecycleAction.isEmpty()) {
```
harsha89: use string utils

modules/api-import-export/src/main/java/org/wso2/carbon/apimgt/importexport/utils/APIImportUtil.java
```diff
@@ -484,7 +644,14 @@ public static void updateAPI(String apiID, String pathToArchive, String currentU
                 }
             }
             // update the API
+            importedApi.setStatus(currentStatus);
             provider.updateAPI(importedApi);
+
+            // Change lifecycle if state transition is required
+            if (lifecycleAction != null && !lifecycleAction.isEmpty()) {
```
harsha89: use stringutils

kasvith:
Resolved, please check

PR #5280, opened 2019-08-08
Remove rest.api.util jar from web-apps lib directory

npamudika:
This PR removes the org.wso2.carbon.apimgt.rest.api.util jar from each web-app's lib directory and adds the same jar into /lib/runtimes/cxf3/ directory, so that it will be picked in the runtime.

Relevant PRs - https://github.com/wso2/carbon-apimgt/pull/6793

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1151,7 +1151,7 @@
         <carbon.analytics.common.version>5.2.1</carbon.analytics.common.version>
 
         <!-- APIM Component Version -->
-        <carbon.apimgt.version>6.5.56-SNAPSHOT</carbon.apimgt.version>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -1151,7 +1151,7 @@
         <carbon.analytics.common.version>5.2.1</carbon.analytics.common.version>
 
         <!-- APIM Component Version -->
-        <carbon.apimgt.version>6.5.34</carbon.apimgt.version>
+        <carbon.apimgt.version>6.5.56-SNAPSHOT</carbon.apimgt.version>
```
malinthaprasan: I think we can't mention the snapshot version due to auto-release

npamudika: Fixed

PR #5440, opened 2019-08-30
Add tls keys to support k8s deployments

kasunbg:
Now, the product scenario tests can be run against k8s deployments, in addition AWS deployments that we already support. 

K8S deployments are done either via plain k8s yamls or via Helm. Both are supported by Testgrid. Google's GKE is used as the infrastructure provider, and we have created a k8s cluster there - dev-test-cluster for dev - https://console.cloud.google.com/kubernetes/list?project=testgrid&authuser=0

To access the APIM deployment @ k8s over HTTPS, we need to add TLS public keys into tests. This PR adds that.


product-scenarios/scenarios-common/src/main/java/org/wso2/am/scenario/test/common/ScenarioTestBase.java
```diff
@@ -162,7 +162,7 @@ public static TenantManagementServiceClient getTenantManagementServiceClient() t
         try {
             authenticatorClient = new AuthenticatorClient(keyManagerURL);
             URL url = new URL(keyManagerURL);
-            String sessionCookie = authenticatorClient.login("admin", "admin", url.getHost());
+            String sessionCookie = authenticatorClient.login("admin", "admin", "localhost");
```
praminda: Why do we need to make this `localhost`? This means we no longer take the property values coming from Deployment Properties.

kasunbg: AuthenticatorClient is already initialized with the correct keyManagerURL as you see two line above.

`new AuthenticatorClient(keyManagerURL);`

The third parameter in #login is not really used as the endpoint. It was used for some unnecessary validations for this case.

kasunbg: AWS deployment tests are passing - https://testgrid-live-dev.private.wso2.com/kasung-wso2am-2.6.0-aws-new-tests

K8S tests pass as well (this has test failures with mysql db due to other reasons) - https://testgrid-live-dev.private.wso2.com/wum-sce-test-helm-wso2am-2.6.0-full

product-scenarios/scenarios-common/src/main/java/org/wso2/am/scenario/test/common/ScenarioTestBase.java
```diff
@@ -162,7 +162,7 @@ public static TenantManagementServiceClient getTenantManagementServiceClient() t
         try {
             authenticatorClient = new AuthenticatorClient(keyManagerURL);
             URL url = new URL(keyManagerURL);
```
praminda: `url` object no longer has any usages
```suggestion
```

product-scenarios/scenarios-common/src/main/java/org/wso2/am/scenario/test/common/ScenarioTestBase.java
```diff
@@ -203,7 +203,7 @@ private String login(String host, String username, String password) throws APIMa
         try {
             authenticatorClient = new AuthenticatorClient(host);
             URL url = new URL(host);
```
praminda: `url` object no longer has any usages
```suggestion
```

Issue #5524, opened 2019-09-11
Adding application attirbutes configurations to deployment.toml


vithu30:
We need to add configurations in deployment.toml to add custom application attributes

PR #5554, opened 2019-09-12
Add application attributes sample config to deployment.toml

tharikaGitHub:
This PR adds a sample configuration for application attributes.

Fixes https://github.com/wso2/product-apim/issues/5524


tharindu1st:
@tharikaGitHub  please add this config as config docs .since we maintain minimal config on deployment.toml we don't need to add it to deployment.toml.

tharikaGitHub:
Closed as per the above comment

PR #5562, opened 2019-09-15
Fix intermittent failure of testWebSocketAPIThrottling

rmsamitha:
Prevents API requests getting dispersed into two units of time in the throttling test. If throttling testing time duration is dispersed into two separate units of time, repeat the test(calling testThrottling method).

### DIFF
--- a/modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/am/integration/tests/websocket/WebSocketAPITestCase.java
+++ b/modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/am/integration/tests/websocket/WebSocketAPITestCase.java
@@ -66,6 +66,7 @@ import java.net.InetAddress;
 import java.net.Socket;
 import java.net.URI;
 import java.net.URL;
+import java.time.LocalDateTime;
 import java.util.List;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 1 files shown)


modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/am/integration/tests/websocket/WebSocketAPITestCase.java
```diff
@@ -356,6 +357,14 @@ public void configure(WebSocketServletFactory factory) {
      * @param accessToken API accessToken
      */
     private void testThrottling(String accessToken) throws Exception {
+
+        final int UNIT_TIME = 5; //5 minutes (The unit time is defined as 5 minutes in policy.json)
+        /* Prevent API requests getting dispersed into two time units */
+        while (LocalDateTime.now().getMinute() % UNIT_TIME >= (UNIT_TIME - 1)) {
```
Rajith90: Isn't this condition reduces to == , rather than >=

rmsamitha: The functionality would be the same if we use '=='
But '==' doesn't cover that the values less than that should return false. In the sense of logic, it is more appropriate to use >= AFAIU , hence code readability would be better.

PR #5915, opened 2019-09-27
Add first Git Action job

bhathiya:


.github/workflows/maven.yml
```diff
@@ -0,0 +1,19 @@
...
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
```
tharindu1st: can we make this 11

bhathiya: planning to add more jobs to cover all combinations. let's start with 8.

PR #5957, opened 2019-09-28
Optimize WebSocket Throttling test

bhathiya:


bhathiya:
I haven't seen any other failing throttling tests. If there are any, we can do the same.

PR #6072, opened 2019-10-01
Fix profiles

bhathiya:


modules/distribution/product/src/main/resources/conf/default.json
```diff
@@ -1,12 +1,13 @@
...
-  "oauth.extensions.token_generator": "org.wso2.carbon.apimgt.keymgt.issuers.APIMTokenIssuer",
   "server.base_path,": "${carbon.protocol}://${carbon.host}:${carbon.management.port}",
   "server.default_cache_timeout": "15",
   "default_cache_timeout.force_local_cache": false,
   "server.enableMTOM": false,
   "server.enableSwA": false,
   "server.discard_empty_caches": false,
   "identity.data_source": "jdbc/WSO2AM_DB",
+  "database.apim_db.url": "jdbc:h2:./repository/database/WSO2AM_DB;DB_CLOSE_ON_EXIT=FALSE;LOCK_TIMEOUT=60000;MVCC=TRUE",
+  "database.apim_db.driver": "org.h2.Driver",
```
ruks: I think this need to be done for "database.apim_db.type" in infer.json

tharindu1st: since this was default.json we not able to do it by that property.

bhathiya: yes, I tried this and it didn't work.

PR #6120, opened 2019-10-02
Fix 'Failed to load class org.slf4j.impl.StaticLoggerBinder' error

bhathiya:


bhathiya:
Note: Ideally this jar should come from the Kernel.

PR #6251, opened 2019-10-04
Migrate test APIM710AllSubscriptionsByApplicationTestCase

Krishanx92:


modules/integration/tests-common/integration-test-utils/src/main/java/org/wso2/am/integration/test/impl/RestAPIStoreImpl.java
```diff
@@ -215,6 +215,13 @@ public HttpResponse createSubscription(String apiId, String applicationId, Strin
         return null;
     }
 
+    public SubscriptionListDTO getSubscription(String apiId, String applicationId, String apiType, String groupId) throws ApiException {
```
ruks: seems length is exceeded.

PR #6355, opened 2019-10-10
change api-store profile to api-devportal

isharac:
changes required to rename api-store profile to api-devportal


modules/distribution/product/src/main/startup-scripts/profileSetup.bat
```diff
@@ -141,7 +141,7 @@ for /f %%i in ('dir %pathToWebapps% /A:-D /b ^| findstr /v "api#am#store#v.*war"
 	endlocal
 )
 rem ---removing jaggeryapps which are not required for this profile--------
-for /f %%i in ('dir "%pathToJaggeryapps%" /A:D /b ^| findstr /v "store"') do (
+for /f %%i in ('dir "%pathToJaggeryapps%" /A:D /b ^| findstr /v "devportal"') do (
```
bhathiya: is this correct?

isharac: the store app is now renamed to devportal

PR #6536, opened 2019-10-19
Add AUTO_SERVER=TRUE to AM_DB

RAVEENSR:
If this section is missing when this AM_DB is accessed through
analytics-apim distribution an error will be given that this db is
locked.


RAVEENSR:
@fazlan-nazeem and @tharindu1st, please review this PR.

PR #6582, opened 2019-10-21
Rename api-store.toml to api-devportal.toml

pubudu538:


bhathiya:
Make sure file copy commands are not broken by this.

PR #6958, opened 2019-11-26
Add Test Case for JWT and LoadBalanceEndpoint

shilmyhasan:
- Includes the test case for JWT


modules/integration/tests-common/integration-test-utils/src/main/java/org/wso2/am/integration/test/impl/RestAPIPublisherImpl.java
```diff
@@ -307,6 +307,16 @@ public void deprecateAPI(String apiId) throws ApiException {
         apiLifecycleApi.apisChangeLifecyclePost(Constants.DEPRECATE, apiId, null, null);
     }
 
+    /**
+     * This method is used to  deploy the api as a prototype.
+     *
+     * @param apiId API id that need to be prototyped.
+     * @throws ApiException throws if an error occurred when publishing the API.
+     */
+    public void prototypeAPI(String apiId) throws ApiException {
```
mushthaq33: Shall we change the method name to a more meaningful one? May be like deployPrototypeAPI().

shilmyhasan: Changed in 768f86bb8f80b599a3b012b566c350648bb4c3e3

PR #6974, opened 2019-11-27
Migrating APIM638ValidateTheRoleOfAnExistingUserThroughThePublisherRestAPITestCase

Meruja:
Migrating APIM638ValidateTheRoleOfAnExistingUserThroughThePublisherRestAPITestCase

### DIFF
--- a/modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/am/integration/tests/publisher/APIM638ValidateTheRoleOfAnExistingUserThroughThePublisherRestAPITestCase.java
+++ b/modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/am/integration/tests/publisher/APIM638ValidateTheRoleOfAnExistingUserThroughThePublisherRestAPITestCase.java
@@ -20,19 +20,19 @@
 
 package org.wso2.am.integration.tests.publisher;
 
-import org.json.JSONObject;
+import org.apache.http.HttpStatus;
 import org.testng.annotations.AfterClass;
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Factory;
 import org.testng.annotations.Test;
+import org.wso2.am.integration.clients.publisher.api.ApiException;
+import org.wso2.am.integration.clients.publisher.api.ApiResponse;
 import org.wso2.am.integration.test.utils.base.APIMIntegrationBaseTest;
-import org.wso2.am.integration.test.utils.clients.APIPublisherRestClient;
 import org.wso2.carbon.automation.engine.context.TestUserMode;
 import org.wso2.carbon.integration.common.admin.client.UserManagementClient;
 
-import static org.testng.Assert.assertFalse;
-import static org.testng.Assert.assertTrue;
+import static org.testng.Assert.*;
 
 /**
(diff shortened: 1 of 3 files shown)


modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/am/integration/tests/publisher/APIM638ValidateTheRoleOfAnExistingUserThroughThePublisherRestAPITestCase.java
```diff
@@ -82,46 +81,32 @@ public void setEnvironment() throws Exception {
...
-                "Invalid response for the Role - " + creatorRole);
-
+        apiResponse = restAPIPublisher.validateRoles(creatorRole);
+        assertEquals(apiResponse.getStatusCode(), HttpStatus.SC_OK,
+                "Response code mismatch for: " + creatorRole);
     }
 
     @Test(groups = {"wso2.am"}, description = "Validate the Role of an non-existing user through" +
-            " the publisher rest API ")
+            " the publisher rest API ", dependsOnMethods = {"testValidateTheRoleOfAnExistingUser"})
```
prasa7, line 101: since the dependsOnMethods only has one method, you can remove the curly braces("{" and "}")

Meruja: Fixed with cf2570e

modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/am/integration/tests/publisher/APIM638ValidateTheRoleOfAnExistingUserThroughThePublisherRestAPITestCase.java
```diff
@@ -82,46 +81,32 @@ public void setEnvironment() throws Exception {
...
-
         //Validate the publisher role
-        JSONObject jsonObjectPublisher = new JSONObject
-                (apiPublisher.validateRoles(publisherRole).getData());
-        assertFalse(jsonObjectPublisher.getBoolean("error"), "Invalid Role Name - " + publisherRole);
-        assertFalse(jsonObjectPublisher.getBoolean("response"),
-                "Invalid response for the Role - " + publisherRole);
-
+        try {
+            restAPIPublisher.validateRoles(publisherRole);
```
uvindra, line 105: Since the expectation of calling this method for a non existing user is to throw an exception, can we add an `Assert.fail()` below this line? It will look like below,

```
try {
            restAPIPublisher.validateRoles(publisherRole);
           Assert.fail("Exception was not thrown when validating role for non existing user");
} catch (ApiException e) {

}
```

In case exception is not thrown as expected the test will fail as well then.

Meruja: Fixed with https://github.com/wso2/product-apim/pull/8816

Issue #7414, opened 2020-02-06
APIM Key Manager Config model issues


tgtshanika:
### Description:
Following configurations has to be removed from IS as KM deployment.toml

1. Config db configuration

```toml
[database.config]
type = "h2"
url ="jdbc:h2:./repository/database/WSO2SHARED_DB;DB_CLOSE_ON_EXIT=FALSE;MVCC=TRUE"
username = "wso2carbon"
password = "wso2carbon"
```

2. TM endpoints

```toml
[apim.throttling]
receiver_url = "tcp://localhost:9611"
receiver_auth_url = "ssl://localhost:9711"
```

tgtshanika:
Tagging https://github.com/wso2/product-apim/issues/7228

PR #7443, opened 2020-02-12
fix is as km related issues

tharindu1st:


tharindu1st:
Fix https://github.com/wso2/product-apim/issues/7414

PR #7929, opened 2020-04-21
Fix issues in migration DB scripts

npamudika:


modules/distribution/resources/migration-2.6.0_to_3.0.0/oracle.sql
```diff
@@ -10,7 +10,7 @@ CREATE TABLE AM_SYSTEM_APPS (
 )
 /
 
-CREATE SEQUENCE AM_SYSTEM_APPS_SEQ START WITH 1 INCREMENT BY 1 NOCACHE
+CREATE AM_SYSTEM_APP_SEQUENCE START WITH 1 INCREMENT BY 1 NOCACHE
```
vithu30, line 13: The query seems to be correct `CREATE SEQUENCE AM_SYSTEM_APP_SEQUENCE START WITH 1 INCREMENT BY 1 NOCACHE`. The one we fixed in docs is different case - https://github.com/wso2/product-apim/pull/7013/. This is fixed in this script with - https://github.com/wso2/docs-apim/pull/1041. Is it for fixing a different issues?

PR #8439, opened 2020-06-24
property added for reading accept header in script mediator

VinulaUthsara:
This allows accept header types to be read from the script mediator in prototype implementation (inline scripts).

Screenshot from 2020-06-24 08-19-50


modules/distribution/resources/api_templates/prototype_template.xml
```diff
@@ -213,6 +213,7 @@ url-mapping="$resource.getUriTemplate()"
 methods="$resource.getMethodsAsString()" faultSequence=#print_string_if_exist($faultSequence "fault")>
 <inSequence>
     #if(!$resource.getMediationScript().equalsIgnoreCase("null"))
+            <property name="AcceptHeader" expression="$trp:Accept"/>
```
malinthaprasan, line 216: shall we fix the indentation here?

PR #8473, opened 2020-06-26
Upadted humantask zip files for WS workflow executor

SahanHerath:
updated human-task zip files to approve/reject WS workflow requests through human-task-explorer. WS requests of application creation, subscription creation, registration creation, user self sign up can be approved through human-task-explorer. API state change WS requests can be approved through BPMN-explorer like the past.


chamilaadhi:
lets format the xml elements inside the .ht files

PR #8695, opened 2020-07-16
Upgrading the versions in carbon.apimgt and wso2iskm

HiranyaKavishani:


HiranyaKavishani:
Close due to conflicts

PR #9006, opened 2020-08-10
remove unused jars

isharac:


modules/distribution/product/src/main/assembly/bin.xml
```diff
@@ -99,6 +99,8 @@
                 <!--<exclude>**/repository/components/plugins/poi_3.9.0.wso2v1.jar</exclude>-->
 
                 <!-- Remove jars which are not needed-->
+                <exclude>**/repository/components/plugins/org.wso2.carbon.feature.mgt.stub_4.6.0.jar</exclude>
+                <exclude>**/repository/components/plugins/org.wso2.carbon.feature.mgt.services_4.6.0.jar</exclude>
                 <exclude>**/repository/components/plugins/hadoop-client_2.7.2.wso2v1.jar</exclude>
```
chamilaadhi, line 104: shall we use ${carbon.kernel.version} as the version

isharac: Fixed with 4c9ad3d1acd6469c9f9ccf9b88de4c3431ef6723
Thanks.

chamilaadhi (approved):

PR #9894, opened 2021-02-09
Add new config to analytics

ruks:
- Adding auth service-related config to analytics

### DIFF
--- a/modules/distribution/product/src/main/conf/deployment.toml
+++ b/modules/distribution/product/src/main/conf/deployment.toml
@@ -92,12 +92,9 @@ https_endpoint = "https://localhost:${https.nio.port}"
 
 [apim.analytics]
 enable = false
-#reporter_impl="class.name"
-#handler_impl="class.name"
-#properties.'auth.api.url'="https://localhost:9443/auth"
-#properties.'token.api.url'="https://localhost:9443/token"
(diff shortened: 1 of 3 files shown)


modules/distribution/product/src/main/conf/deployment.toml
```diff
@@ -92,12 +92,9 @@ https_endpoint = "https://localhost:${https.nio.port}"
...
 enable = false
-#reporter_impl="class.name"
-#handler_impl="class.name"
-#properties.'auth.api.url'="https://localhost:9443/auth"
-#properties.'token.api.url'="https://localhost:9443/token"
-#properties.'consumer.key'="consumer_key"
-#properties.'consumer.secret'="consumer_secret"
+config_endpoint = "https://localhost:8080/auth/v1"
+auth_token = ""
+properties.'worker.thread.count' = 1
```
jaadds, line 97: Why do we need to have it as a property?

jaadds (approved):

modules/distribution/product/src/main/conf/deployment.toml
ruks: the event not publishing issue is due to the multi-threads. So as a temporary solution I'm limiting to single worker threads,

PR #10315, opened 2021-03-17
adding internal-key cors header

hasuniea:


modules/distribution/product/src/main/conf/deployment.toml
```diff
@@ -146,7 +146,7 @@ properties.'worker.thread.count' = 1
 [apim.cors]
 allow_origins = "*"
 allow_methods = ["GET","PUT","POST","DELETE","PATCH","OPTIONS"]
-allow_headers = ["authorization","Access-Control-Allow-Origin","Content-Type","SOAPAction","apikey", "testKey"]
+allow_headers = ["authorization","Access-Control-Allow-Origin","Content-Type","SOAPAction","apikey","Internal-Key"]
```
tmkasun, line 149: Do we need to allow this header anymore, since we use the `bearer` token for internal testing as well?

tmkasun: sorry i think it's `Internal-Key` header not the regular authorization header

tmkasun (approved):

modules/distribution/product/src/main/conf/deployment.toml
hasuniea: No, as I discussed with Tharindu, we can't use the Authorization header. We have to use the Internal-Key header so that anybody can write a handler and add their own logic to authenticate the request.

Issue #10341, opened 2021-03-18
Add HTTPS support for Webhooks APIs


chanikag:
### Describe your problem(s)
The webhooks provider should be able to send events over HTTPS.

### How will you implement it

---

PR #10342, opened 2021-03-18
Add HTTPS webhooks inbound endpoint

chanikag:
This is to fix https://github.com/wso2/product-apim/issues/10341


modules/distribution/product/src/main/conf/synapse-configs/default/inbound-endpoints/SecureWebhookServer.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<inboundEndpoint name="SecureWebhookServer"
+                 protocol="https"
+                 suspend="false" sequence="TestIn" onError="fault" >
+    <p:parameters xmlns:p="http://ws.apache.org/ns/synapse">
+        <p:parameter  name="inbound.http.port">8021</p:parameter>
+        <p:parameter name="keystore">
+            <KeyStore>
+                <Location>repository/resources/security/wso2carbon.jks</Location>
```
tharindu1st, line 9: if this key stores are hardcoded will become impact on changing this can we add templated file

tharindu1st (changes requested):

modules/distribution/product/src/main/conf/synapse-configs/default/inbound-endpoints/SecureWebhookServer.xml
chanikag: +1. But we have done the same for other inbound endpoints.

tharindu1st: yes

tharindu1st (approved):

PR #10370, opened 2021-03-18
Integration testing for Pagination

thisaltennakoon:


modules/integration/tests-common/integration-test-utils/src/main/java/org/wso2/am/integration/test/impl/RestAPIPublisherImpl.java
```diff
@@ -1824,7 +1824,9 @@ public HttpResponse editComment(String commentId, String apiId, String comment,
 
             CommentDTO editedCommentDTO = commentsApi.editCommentOfAPI(commentId,apiId,patchRequestBodyDTO);
             if (editedCommentDTO != null) {
-                response = new HttpResponse(gson.toJson(editedCommentDTO), 201);
+                response = new HttpResponse(gson.toJson(editedCommentDTO), 200);
+            } else {
+                response = new HttpResponse(null, 200);
```
malinthaprasan, line 1829: Which case do we get a response with empty body here with 200?

thisaltennakoon: When we sending an edit request without doing any changes. Earlier, NOT MODIFIED was the response but it has been changed to 200

malinthaprasan (approved):

PR #10466, opened 2021-03-28
update carbon.mediation.version

tharindu1st:
update synapse.version
update testcase to support portoffset


Nirothipan (approved):

Nirothipan:
@tharindu1st shall we remvoe .pom.xml.swp file

tharindu1st:
> @tharindu1st shall we remvoe .pom.xml.swp file
done

Issue #9773, opened 2021-01-22
URI Template is not passed when default version is called with SOAPAction header (Backend is SOAP endpoint)


sajinidesilva:
### Description:
If you call the default version of API, where API has a soap backend, with SOAPAction header the URI temple will not be appended to the backed call

### Steps to reproduce:

1. Create an API with soap endpoint.
2. Enable default versioning
3. Publish the API.
4. Send a request to the API without the version. Should add "SOAPAction" header when sending the request

### Affected Product Version:
2.5.0

### Environment details (with versions):
- OS:
- Client:
- Env (Docker/K8s):

---

msm1992:
Resolving this as the default api template is not used in master as stated in https://github.com/wso2/product-apim/pull/10827

PR #10827, opened 2021-04-08
Set default api message format to REST

msm1992:
Fixes wso2/product-apim#9773


tharindu1st:
default api template not used in 4.0.0 release please remove the template since we no longer use this

PR #10975, opened 2021-04-12
add control-plane configs and profile

tharindu1st:


modules/distribution/product/src/main/resources/conf/infer.json
```diff
@@ -137,5 +147,13 @@
       "synapse_handlers.analytics_handler.name": "analytics_handler",
       "synapse_handlers.analytics_handler.class": "$ref{apim.analytics.handler_impl}"
     }
+  },
+  "server.disable_traffic_manager": {
```
bhathiya, line 151: This should be "enable". Shall we use "management"? 

so it will be `enable_traffic_management`. wdyt?

modules/distribution/product/src/main/resources/conf/infer.json
```diff
@@ -137,5 +148,13 @@
       "synapse_handlers.analytics_handler.name": "analytics_handler",
       "synapse_handlers.analytics_handler.class": "$ref{apim.analytics.handler_impl}"
     }
+  },
+  "server.enable_traffic_management": {
+    "true": {
```
bhathiya, line 153:
```suggestion
    "false": {
```

bhathiya (changes requested):

bhathiya (approved):

PR #10983, opened 2021-04-12
Upgrade dependencies

sajinieKavindya:
This PR upgrades mediation, synapse, and axis2-transport versions


Arshardh:
We encountered a build failure due to axis2 transport version upgrade, shall we demote that version back to the original

malinthaprasan (approved):

PR #11172, opened 2021-05-04
fixes #11141

### LINKED ISSUE #11141: Analytics event publisher url should be updated
### Describe your problem(s)
- ATM analytics publishing endpoint is not defined in the default config and needs to refer to the docs and apply the endpoint. 

### Describe your solution
- Since this prod endpoint is not changed we don't need to configure it manually. Instead, the product can have the default configuration pointing to the prod.

### How will you implement it
From the deployment.toml we can remove the 'config_endpoint' key and define it in the default.json. If this endpoint is changed or needs to point to a different endpoint, they can override it in the deployment.toml

hasuniea:


modules/distribution/product/src/main/resources/conf/default.json
```diff
@@ -338,6 +338,7 @@
   "qpid.queue.auto_register": true,
   "qpid.queue.view_message_counts": false,
   "qpid.status_updates": "ON",
+  "config_endpoint": "https://localhost:8080/auth/v1",
```
ruks, line 341: we can give the actual prod endpoint now. We can use this endpoint now. 
```
"config_endpoint": "https://analytics-event-auth.choreo.dev/auth/v1"
```

ruks (approved):

Issue #11270, opened 2021-05-28
Issue when response cache is enabled for an API


Manasha11:
### Description:
When response cache is enabled for an API via the UI, even if the API is invoked multiple times during the timeout period, the events do not get published to Analytics. 
Hence the correct number of hits for the API does get reflected in the Usage Summary widget of APIM Publisher dashboard (the request count is only increased by 1).

### Steps to reproduce:
Create an API and enable response cache.
Invoke the API multiple times during defined timeout period.
View the usage stats in the Analytics dashboard.

### Affected Product Version:
3.1.0, 3.2.0

isharac:
need to add below in the insequence of the api.

    <onCacheHit>
        <class name="org.wso2.carbon.apimgt.gateway.handlers.analytics.APIMgtResponseHandler"/>
        <respond/>
    </onCacheHit>

npamudika:
Fixed with https://github.com/wso2/product-apim/pull/11323

PR #11323, opened 2021-06-11
fix https://github.com/wso2/product-apim/issues/11270

isharac:


isharac:
Added with https://github.com/wso2/product-apim/pull/11377

Issue #11435, opened 2021-07-15
Issue in API creation with OAuth secured endpoints via Publisher REST API


chaliya96:
### Description:
When creating an API with oauth secured endpoints in the publisher REST API (POST resource), the following payload will not create secured endpoint but it will create a prototype endpoint.  
`
{
    "name": "api-new",
    "version": "1",
    "context": "/test99",
    "policies": [
        "Unlimited"
    ],
    "endpointConfig": {
        "endpoint_type": "http",
        "sandbox_endpoints": {
            "url": "http://localhost:3001/test/two"
        },
        "production_endpoints": {
            "url": "http://localhost:3001/test/two"
        },
                "endpoint_security": {
            "sandbox": {
                "password": null,
                "tokenUrl": "http://localhost:9443/token",
                "clientId": "cid123",
                "clientSecret": "csecret",
                "customParameters": "{}",
                "type": "OAUTH",
                "grantType": "CLIENT_CREDENTIALS",
                "enabled": true,
                "username": null
            },
            "production": {
                "password": null,
                "tokenUrl": "http://localhost:9443/token",
                "clientId": "cid123",
                "clientSecret": "csecret",
                "customParameters": "{}",
                "type": "OAUTH",
                "grantType": "CLIENT_CREDENTIALS",
                "enabled": true,
                "username": null
            }
        }
    },
    "gatewayEnvironments": [
        "Production and Sandbox"
    ]
}
`
But with the PUT resource same payload will update the endpoint to the secured endpoint. 

### Steps to reproduce:
Generate valid access token for Publisher REST API - api creation. 
Invoke the API using sample payload

### Affected Product Version:
v3.2.0

wasuradananjith:
An error can be observed as below when performing the above operation.

```
[2021-07-16 10:17:06,258] ERROR - APIMappingUtil Error while decrypting client credentials for API: admin-PizzaShackAPI-1.0.0
org.wso2.carbon.core.util.CryptoException: An error occurred while decrypting data.
	at org.wso2.carbon.core.util.CryptoUtil.decrypt(CryptoUtil.java:265) ~[org.wso2.carbon.core_4.6.0.jar:?]
	at org.wso2.carbon.core.util.CryptoUtil.base64DecodeAndDecrypt(CryptoUtil.java:329) ~[org.wso2.carbon.core_4.6.0.jar:?]
[...]

PR #11444, opened 2021-07-19
Adding the test cases for adding endpoint security during /apis: POST

wasuradananjith:
## Purpose
Fixes: https://github.com/wso2/product-apim/issues/11435

## Goals
Adding the integration tests to add an API with different types of endpoint security which were not covered before.

## Approach
- Added a new file named **AddEndPointSecurityPerTypeTestCase.java** with the relevant test cases.

## Related PRs
- https://github.com/wso2/carbon-apimgt/pull/10684

## Test environment
- JDK 1.8.0_251
- OS: Ubuntu 20.04.2 LTS

### DIFF
--- a/modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/am/integration/tests/api/lifecycle/AddEndPointSecurityPerTypeTestCase.java
+++ b/modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/am/integration/tests/api/lifecycle/AddEndPointSecurityPerTypeTestCase.java
@@ -1,0 +1,475 @@
+/*
+ * Copyright (c) 2021, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ * WSO2 Inc. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.wso2.am.integration.tests.api.lifecycle;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.http.HttpEntity;
+import org.apache.http.client.entity.UrlEncodedFormEntity;
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpPost;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClientBuilder;
+import org.apache.http.message.BasicNameValuePair;
+import org.json.simple.JSONObject;
+import org.json.simple.parser.JSONParser;
+import org.json.simple.parser.ParseException;
+import org.testng.Assert;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Factory;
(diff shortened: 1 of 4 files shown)


modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/am/integration/tests/api/lifecycle/AddEndPointSecurityPerTypeTestCase.java
```diff
@@ -0,0 +1,475 @@
...
+        httpPost.setEntity(urlEncodedFormEntity);
+        try (CloseableHttpResponse response = closeableHttpClient.execute(httpPost)) {
+            Assert.assertEquals(response.getStatusLine().getStatusCode(), 200);
+            HttpEntity entity = response.getEntity();
+            JSONObject jsonPayload = (JSONObject) new JSONParser().parse(new InputStreamReader(entity.getContent()));
+            Assert.assertTrue((Boolean) jsonPayload.get("active"));
+            Assert.assertNotNull(jsonPayload.get("client_id"));
+            Assert.assertEquals(jsonPayload.get("client_id"), clientId);
+        } catch (IOException | ParseException e) {
+            e.printStackTrace();
```
tharikaGitHub, line 472: Here instead of printing the entire stack trace, shall we catch ad throw the exception and print only the message here?

tharikaGitHub (approved):

modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/am/integration/tests/api/lifecycle/AddEndPointSecurityPerTypeTestCase.java
wasuradananjith: Fixed via https://github.com/wso2/product-apim/pull/11444/commits/eb2a3225ca45b153d419d20862f9affb6874b92a

modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/am/integration/tests/api/lifecycle/AddEndPointSecurityPerTypeTestCase.java
```diff
@@ -0,0 +1,476 @@
...
+        // Create an API
+        String apiName = API_NAME + "1";
+        String apiContext = API_CONTEXT + "1";
+        APICreationRequestBean apiCreationRequestBean =
+                new APICreationRequestBean(apiName, apiContext, API_VERSION_1_0_0, providerName,
+                        new URL(apiEndPointUrl), new URL(apiEndPointUrl));
+        apiCreationRequestBean.setTier(TIER_UNLIMITED);
+        apiCreationRequestBean.setTiersCollection(TIER_UNLIMITED);
+        apiCreationRequestBean.setProvider(user.getUserName());
+        apiCreationRequestBean.setSetEndpointSecurityDirectlyToEndpoint(Boolean.TRUE);
```
malinthaprasan, line 149: Is setting this FALSE (or not setting it) already handled in other test cases?

malinthaprasan (approved):

modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/am/integration/tests/api/lifecycle/AddEndPointSecurityPerTypeTestCase.java
wasuradananjith: Yes, the only place this function will be used is inside this test case file by setting this as TRUE. 
By default this value **setSetEndpointSecurityDirectlyToEndpoint** will have FALSE [1]. That means, in every other place this will be FALSE so that the else part of the if statement in [2] will get executed which aligns with the original behaviour we had before adding this new parameter.

[1] https://github.com/wso2/product-apim/pull/11444/files#diff-987c398ba7ceb48bac9e57af64a5953d3dad4dc95edbb5bdc44ac2c3a43b8c2fR110
[2] https://github.com/wso2/product-apim/pull/11444/files#diff-e2b0209eebfea247bd6a651abdc6ce6f47f32fa2e00f311fa7a34d76cd0404c5R1186

Issue #11894, opened 2021-10-22
AUD and ISS values are missing from the Introspect response of the APIM 3.2.0


AshenJayasuriya:
### Description:
AUD and ISS values are missing from the Introspect response of the APIM 3.2.0 where the same values are present in the Introspect response of the APIM 2.6.0.

### Steps to reproduce:

- Login to the devportal.
- Create an application.
- Generate keys and get an access token.
- Introspect the access token by calling the introspect endpoint.

The response will be as below.

```
{
    "nbf":1633686737,
    "scope":"default",
    "active":true,
    "token_type":"Bearer",
    "exp":1633690337,
    "iat":1633686737,
    "client_id":"71lKp7Y4Rn38emzfOKqy1iHDFkwa",
    "username":"admin@carbon.super"
}
```

### Suggested solution:

Add the missing AUD and ISS values to the introspect response.

### Affected Product Version:
3.2.0

PR #11987, opened 2021-11-11
Provide APIM support for JWT token validation

YasasRangika:
### Purpose
To provide support for JWT type-token validation

### Goal
Fixes: https://github.com/wso2/product-apim/issues/11894

### Approach
Introduced 
```
[oauth]
enable_jwt_token_validation_during_introspection = true
```
to enable/disable JWT token validation


tharindu1st:
This config no need to put it into deployment.toml
since this already in config catalogs

Issue #12085, opened 2021-11-30
Error starting gateway-worker, traffic-manager profile


dushaniw:
### Description:
getting following error for gateway-worker and traffic-manager profiles

```
[2021-11-30 16:40:29,599] ERROR - [/] Exception starting filter [CaptchaFilter]
java.lang.ClassNotFoundException: org.wso2.carbon.identity.captcha.filter.CaptchaFilter
	at org.apache.catalina.loader.WebappClassLoaderBase.loadClass(WebappClassLoaderBase.java:1407) ~[tomcat_9.0.52.wso2v2.jar:?]
	at org.apache.catalina.loader.WebappClassLoaderBase.loadClass(WebappClassLoaderBase.java:1215) ~[tomcat_9.0.52.wso2v2.jar:?]
	at org.apache.catalina.core.DefaultInstanceManager.loadClass(DefaultInstanceManager.java:538) ~[tomcat_9.0.52.wso2v2.jar:?]
	at org.apache.catalina.core.DefaultInstanceManager.loadClassMaybePrivileged(DefaultInstanceManager.java:519) ~[tomcat_9.0.52.wso2v2.jar:?]
	at org.apache.catalina.core.DefaultInstanceManager.newInstance(DefaultInstanceManager.java:149) ~[tomcat_9.0.52.wso2v2.jar:?]
	at org.apache.catalina.core.ApplicationFilterConfig.getFilter(ApplicationFilterConfig.java:249) ~[tomcat_9.0.52.wso2v2.jar:?]
	at org.apache.catalina.core.ApplicationFilterConfig.<init>(ApplicationFilterConfig.java:101) ~[tomcat_9.0.52.wso2v2.jar:?]
	at org.apache.catalina.core.StandardContext.filterStart(StandardContext.java:4613) [tomcat_9.0.52.wso2v2.jar:?]
	at org.apache.catalina.core.StandardContext.startInternal(StandardContext.java:5256) [tomcat_9.0.52.wso2v2.jar:?]
	at org.apache.catalina.util.LifecycleBase.start(LifecycleBase.java:183) [tomcat_9.0.52.wso2v2.jar:?]
	at org.apache.catalina.core.ContainerBase$StartChild.call(ContainerBase.java:1396) [tomcat_9.0.52.wso2v2.jar:?]
	at org.apache.catalina.core.ContainerBase$StartChild.call(ContainerBase.java:1386) [tomcat_9.0.52.wso2v2.jar:?]
[...]

PR #12087, opened 2021-11-30
Fixes error starting gateway-worker, traffic-manager profile due to Captcha filter 

Chamindu36:
### Purpose 
Fixes https://github.com/wso2/product-apim/issues/12085 for Master


modules/distribution/product/src/main/resources/conf/templates/repository/conf/tomcat/carbon/WEB-INF/tm.web.xml.j2
```diff
@@ -0,0 +1,221 @@
...
+ <param-value>false</param-value>
+</init-param>
+<init-param>
+<param-name>xssProtectionEnabled</param-name>
+<param-value>false</param-value>
+</init-param>
+</filter>
+
+<filter>
+    <filter-name>CaptchaFilter</filter-name>
```
dushaniw, line 113: dont we need to remove this captcha filter?

rmsamitha (changes requested):
Isn't it better to use the same naming convention we have used for axis2.xml and registry.xml?
i.e. web_TM.xml.j2

dushaniw (approved):

tharindu1st (approved):

rmsamitha (approved):

PR #12344, opened 2022-02-10
[New Feature] Operation Level Policy Support with WSO2 API Manager

ashera96:


modules/distribution/resources/operation_policies/specifications/addQueryParams.json
```diff
@@ -1,7 +1,7 @@
 {
   "category": "Mediation",
-  "name": "addQueryParams",
-  "displayName": "Add a query param",
+  "name": "addQueryParam",
```
1akshitha, line 3: @ashera96 If we change the spec name, we need to update the definition name as well. also this json file's name.

Krishanx92 (approved):

PR #12916, opened 2022-06-17
Generate Service-Catalog API Client with OpenAPI codegen

Sachini-Dissanayaka:
## Purpose
Related to: https://github.com/wso2/api-manager/issues/197

## Goals
Creating service-catalog API client with `openapi-generator-maven-plugin`


modules/integration/tests-common/clients/service-catalog/pom.xml
```diff
@@ -0,0 +1,133 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+    Copyright (c) 2019, WSO2 Inc. (http://wso2.com) All Rights Reserved.
```
vithu30, line 3: Update the year in license header

vithu30:
Since we are adding a new template file with modifications, it will affect the other REST API client as well. Shall we re-generate those clients and commit the changes too (We can send it using a different PR).

Sachini-Dissanayaka:
> Since we are adding a new template file with modifications, it will affect the other REST API client as well. Shall we re-generate those clients and commit the changes too (We can send it using a different PR).

Have added those changes in the "Update service-catalog client" commit

vithu30:
> > Since we are adding a new template file with modifications, it will affect the other REST API client as well. Shall we re-generate those clients and commit the changes too (We can send it using a different PR).
> 
> Have added those changes in the "Update service-catalog client" commit

Seems that commit only has changes of Service catalog client. Since we are using same template files for other clients in [1], those also need to be updated.

[1] https://github.com/wso2/product-apim/tree/master/modules/integration/tests-common/clients

vithu30 (approved):

PR #13003, opened 2022-11-14
Fix Self SignUp Issue on the Response Message

piyumaldk:
Message comes from IS product. Since we are directly registering users when self sign up, our passing prop for IS is wrong. Changed it the appropriate one. 

New Screenshot - 
image

### DIFF
--- a/modules/distribution/product/src/main/extensions/self-registration-complete.jsp
+++ b/modules/distribution/product/src/main/extensions/self-registration-complete.jsp
@@ -107,7 +107,7 @@
             } else {
                 if (isEmailNotificationEnabled) {
             %>
-            <p><%=IdentityManagementEndpointUtil.i18n(recoveryResourceBundle, "Confirmation.sent.to.mail")%>
+            <p><%=IdentityManagementEndpointUtil.i18n(recoveryResourceBundle, "User.registration.completed.successfully")%>
             </p>
             <% } else {%>
             <p><%=IdentityManagementEndpointUtil.i18n(recoveryResourceBundle,


modules/distribution/product/src/main/extensions/self-registration-complete.jsp
```diff
@@ -107,7 +107,7 @@
             } else {
                 if (isEmailNotificationEnabled) {
```
tharikaGitHub, line 108: Seems "isEmailNotificationEnabled" this property is coming in as true. In that case the confirmation sent to mail message is correct. Shall we check where we have enabled this property to true?

piyumaldk: Initially that value is "false".
```
isEmailNotificationEnabled = Boolean.parseBoolean(application.getInitParameter(
            IdentityManagementEndpointConstants.ConfigConstants.ENABLE_EMAIL_NOTIFICATION));
```
Then the value is updated by this. 
_**IdentityManagementEndpointConstants.ConfigConstants.ENABLE_EMAIL_NOTIFICATION**_ is defined as **"EnableEmailNotification"** in IS product [[here](https://github.com/wso2/carbon-identity-framework/blob/31ca11eba35930e0f0214b96946eb2ca0a0dd00f/components/identity-mgt/org.wso2.carbon.identity.mgt.endpoint.util/src/main/java/org/wso2/carbon/identity/mgt/endpoint/util/IdentityManagementEndpointConstants.java#:~:text=public%20static%20final%20String%20ENABLE_EMAIL_NOTIFICATION%20%3D%20%22EnableEmailNotification%22%3B)].

Since this is a constant, and in APIM, we are not sending emails, should I remove this assignment? Then the issue will be solved again.

piyumaldk:
Closing this PR as this will be addressed in a different fix

PR #13048, opened 2022-12-09
Integration tests for PR carbon-apimgt#11710

IsuruMaduranga:
## Purpose

1. Integration tests for https://github.com/wso2-support/carbon-apimgt/pull/5427
2. Fixes https://github.com/wso2/api-manager/issues/760


modules/integration/tests-integration/tests-backend/src/test/resources/oas/v2/oas.json
```diff
@@ -55,7 +55,7 @@
           "default" : [ "SwaggerPetstore_v2" ]
         } ],
         "x-auth-type" : "Application & Application User",
-        "x-throttling-tier" : "10KPerMin"
+        "x-throttling-tier" : "Unlimited"
```
rmsamitha, line 58: Why do we need to change this? Isn't unlimited tier disabling is tested in separate test cases?

IsuruMaduranga: Because, here both resource level and API level throttling tiers are set. In that case resource level throttling get's the priority and that leads to deletion of API level "x-throttling-tier" parameter from the OAS definition. This causes OAS test cases to fail because it compares before and after OAS definitions.

rmsamitha (approved):

PR #13087, opened 2023-01-23
Fix for - Set Jaeger Tracing sampler type

prasa7:


prasa7:
do not need ot add this since the origina fix changed

PR #13097, opened 2023-01-25
Add the http.block_service_list property to APIM

YasasRangika:
### Purpose
To fix adding the following configuration in the deployment.toml has no effect.

[transport.http]
block_service_list = false

### Goal
Fixes: https://github.com/wso2/api-manager/issues/1298

### Instructions
Add the following configuration in deployment.toml to give access to the service list if not **default value is true.**

[transport.http]
block_service_list = false

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1475,7 +1475,7 @@
         <!-- WSDL4J dependencies -->
         <wsdl4j.version>1.6.3.wso2v3</wsdl4j.version>
         <wso2is.km.version>1.7.11</wso2is.km.version>
-        <wso2is7.km.version>2.0.2</wso2is7.km.version>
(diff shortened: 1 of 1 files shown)


YasasRangika:
Marked as do not merge due to an issue in [1]. Properties are always reading from nhttp configs in both PassThroughNHttpGetProcessor.java and DefaultHttpGetProcessor.java classes.

[1] https://github.com/wso2-support/wso2-synapse/blob/69a1633fd0d6453bd150730b2ad123e76c5bf539/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/api/PassThroughNHttpGetProcessor.java#L432

PR #13118, opened 2023-02-01
Disable all types of correlation logs at the end

imsuneth:
Related to https://github.com/wso2/api-manager/issues/1204

Prevent failing Run 2 if Run 1 failed in the middle


rmsamitha:
Note: This is useful only when test failures in CorrelationLoggingTest class occurred causing some tests/methods/sections of methods not to execute.
(i.e. Correlation disabling calls were not executed due to prior failures)

rmsamitha (approved):

PR #13252, opened 2023-07-18
Add testcase for tenant user login with scope assignment

HeshanSudarshana:
This PR adds the testcase for the tenant user login issue when the user is assigned a custom role with only scope assignments.

Related issue: https://github.com/wso2/api-manager/issues/1169


HeshanSudarshana:
Closing the PR since this scenario is tested in other test cases.

PR #13316, opened 2024-01-04
Fix tenantDomain param issue when T is capital in param

chashikajw:
## Purpose
The issue occurs in the JIT provisioning when we choose  the option 'Prompt for username, password, and consent.'
 In the URL build of the JIT provisioning flow, it constructs the URL by adding the capital 'TenantDomain' parameter.

In this fix, we are handling both the TenantDomain and tenantDomain params.

### Fixes
- https://github.com/wso2/api-manager/issues/2344


chashikajw:
This PR is now not needed since this same change has already added from PR https://github.com/wso2/product-apim/commit/70e7c5a977f21c3bbf5d0b9366c3bdea08769706. Hence closing this.

PR #13510, opened 2024-07-08
Add changes to provide support for building the API manager product with JDK 17

YasasRangika:
### Purpose
To make the necessary changes to provide support for building the API manager product with JDK 17.

### Goal
Fixes: https://github.com/wso2/api-manager/issues/2943


YasasRangika:
The task is being closed for now as it is on hold until we decide to start it again.

PR #13517, opened 2024-08-06
Add integration test for audience validation

BiyonFernando:
## Purpose
To add integration tests for audience validation support feature described in [1].

[1] - https://github.com/wso2/api-manager/issues/2784 

## Approach
The `audience` property is included to the schemas in the `publish-api.yaml`. Integration tests are written for the following scenarios to test the functionality of audience validation support:
- Test the response for a GET request without enabling audience validation.
- Test blocking the response when an invalid audience is provided.
- Test whether the response passes through when a valid audience is provided.


RakhithaRR:
Shall we add a test case for the audience validation successful scenario as well? Basically, you can create a new application and generate keys to get the client id and use that value as the audience when creating the API. Then when you subscribe and invoke, the audience validation will happen and return 200 when successful.

We might have to rename the test methods to the successful and failure scenarios.

PR #13730, opened 2025-03-06
Update diagnosis-tool configs to address default config loading issue

YasasRangika:
### Purpose
To resolve the issue of the diagnostic-tool configurations not being loaded properly.

### Goal
Fixes: https://github.com/wso2/api-manager/issues/3595

### Approach

The `repository/resources/conf/default.json` file contains default values for the configurations. Diagnosis-tool configurations are organized into sub-objects, which results in the configuration values provided via `deployment.toml` not working with the config parser. This update flattens the sub-objects to resolve the issue. For eg:

```
"server_configuration": {
        "diagnostic_tool_enabled": "true",
        ...
}
```

will be replaced as:

`"server_configuration.diagnostic_tool_enabled": "true"`

The keys to the following objects have been flattened like this:

- cpu_watcher
- log_watcher
- memory_watcher
- server_configuration
- traffic_analyzer
- zip_file_configuration

-- Additional Improvement:
Introduced the ability to provide the configs with `diagnostic_tool` prefix (without prefix was the previous behaviour, and it is still supported), for better readability.

Eg:
```
[server_configuration]
diagnostic_tool_enabled = "false"
...
```

the above can now be expressed as follows, for better readability:

```
[diagnostic_tool.server_configuration]
diagnostic_tool_enabled = "false"
...
```

Both the above will be supported from this addition.

### DIFF
--- a/all-in-one-apim/modules/distribution/product/src/main/resources/conf/templates/diagnostics-tool/conf/config.toml.j2
+++ b/all-in-one-apim/modules/distribution/product/src/main/resources/conf/templates/diagnostics-tool/conf/config.toml.j2
@@ -22,19 +22,83 @@
 #
 ##############################################
 
+{% macro get_server_configuration(key) %}
+{%- if server_configuration[key] is defined -%}
+{{ server_configuration[key] }}
+{%- elif diagnostic_tool['server_configuration'][key] is defined -%}
+{{ diagnostic_tool['server_configuration'][key] }}
+{%- endif -%}
+{% endmacro %}
+
+{% macro get_cpu_watcher(key) %}
+{%- if cpu_watcher[key] is defined -%}
+{{ cpu_watcher[key] }}
+{%- elif diagnostic_tool['cpu_watcher'][key] is defined -%}
+{{ diagnostic_tool['cpu_watcher'][key] }}
+{%- endif -%}
+
+{% endmacro %}
+{% macro get_memory_watcher(key) %}
+{%- if memory_watcher[key] is defined -%}
+{{ memory_watcher[key] }}
+{%- elif diagnostic_tool['memory_watcher'][key] is defined -%}
+{{ diagnostic_tool['memory_watcher'][key] }}
+{%- endif -%}
+{% endmacro %}
+
+{% macro get_log_watcher(key) %}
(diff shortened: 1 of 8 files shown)


all-in-one-apim/modules/distribution/product/src/main/resources/conf/templates/diagnostics-tool/conf/config.toml.j2
```diff
@@ -22,19 +22,83 @@
...
-server_name = "@product.name@"
-server_version = "@product.version@"
+diagnostic_tool_enabled = "{{ get_server_configuration('diagnostic_tool_enabled') }}"
+deployment_toml_path = "{{ get_server_configuration('deployment_toml_path') }}"
+logs_directory = "{{ get_server_configuration('logs_directory') }}"
+updates_config_path = "{{ get_server_configuration('updates_config_path') }}"
+diagnostic_log_file_path = "{{ get_server_configuration('diagnostic_log_file_path') }}"
+carbon_log_file_path = "{{ get_server_configuration('carbon_log_file_path') }}"
+process_id_path = "{{ get_server_configuration('process_id_path') }}"
+server_name = "WSO2 API Manager"
```
senthuran16: @YasasRangika Will the server name & server version be got via the following ones, similar to what's happening in PR [1] ?
```
server_name = "@product.name@"
server_version = "@product.version@"
```

Since this is a PR for the main branch, I hope it's not correct to hardcode a version here.

[1] https://github.com/wso2-support/product-apim/pull/2303/files#diff-094840970e5b40691fca3b849b28631bdd2181c68df9be4d0fa04b78d6fef68dR100

senthuran16 (changes requested):

all-in-one-apim/modules/distribution/product/src/main/resources/conf/templates/diagnostics-tool/conf/config.toml.j2
YasasRangika: Fixed with https://github.com/wso2/product-apim/pull/13730/commits/f4fc1f27f00e50afb6cbf5aa49ef045b0361d1fa

senthuran16 (approved):

PR #13823, opened 2025-08-26
Add Test for API Policy does not execute when having / as resource

piyumaldk:
$Subject

Related to: https://github.com/wso2/api-manager/issues/4118


piyumaldk:
carbon-apimgt should be 9.32.37 or above to pass the test

PR #13830, opened 2025-09-03
Adding Lua and WASM Policies for KGW APIs

O-sura:
## Purpose
- This PR adds the changes required for introducing Lua and WASM policies to the APIs deployed into KGW. Also for the OperationPolicyTestCase and ELKAnalyticsWithRespondMediatorTestCase tests, a slight modification was added to the getAllCommonOperationPolicies() method, increasing the limit to prevent test failures happening due to exceeding the fetch limit after adding the new policies.

> Related Issues: 
> - https://github.com/wso2/apk/issues/3180

### DIFF
--- a/all-in-one-apim/modules/distribution/resources/operation_policies/specifications/apkWASMInterceptorService_v1.json
+++ b/all-in-one-apim/modules/distribution/resources/operation_policies/specifications/apkWASMInterceptorService_v1.json
@@ -1,0 +1,85 @@
+{
+  "category": "Mediation",
+  "name": "apkWASMInterceptorService",
+  "version": "v1",
+  "displayName": "WASM Extension Service",
+  "description": "This policy allows you to call an WASM extension service for a request message",
+  "policyAttributes": [
+    {
+      "name": "policyName",
+      "displayName": "Policy Name",
+      "description": "Name for the WASM policy(Must be unique within the gateway).",
+      "validationRegex": "^([a-zA-Z_\\:][a-zA-Z\\d_\\-\\ ]*)$",
+      "type": "String",
+      "required": true
+    },
+    {
+      "name": "rootId",
+      "displayName": "Root ID",
+      "description": "This uniquely identifies a set of extensions within a VM that share the same RootContext and Contexts.",
+      "validationRegex": "^([a-zA-Z_\\:][a-zA-Z\\d_\\-\\ ]*)$",
+      "type": "String",
+      "required": true
+    },
+    {
+      "name": "wasmpolicyURL",
+      "displayName": "Policy URL",
+      "description": "The HTTP URL containing the Wasm code.",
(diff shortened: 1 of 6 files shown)


all-in-one-apim/modules/integration/tests-common/integration-test-utils/src/main/java/org/wso2/am/integration/test/impl/RestAPIPublisherImpl.java
```diff
@@ -2439,7 +2439,7 @@ public Map<String, String> getAllCommonOperationPolicies() throws ApiException {
 
         setActivityID();
         ApiResponse<OperationPolicyDataListDTO> apiResponse =
-                operationPoliciesApi.getAllCommonOperationPoliciesWithHttpInfo(50, 0, null);
+                operationPoliciesApi.getAllCommonOperationPoliciesWithHttpInfo(200, 0, null);
```
BLasan, line 2442: Can't we make the default limit to a constant? And make consistent across all places we use that constant. Also, why are we using 200 instead 50? Is this an arbitrary value?

all-in-one-apim/modules/integration/tests-integration/tests-backend/src/test/java/org/wso2/am/integration/tests/analytics/ELKAnalyticsWithRespondMediatorTestCase.java
```diff
@@ -115,7 +115,7 @@ public void testELKAnalyticsWithRespondMediator() throws Exception {
 
         // Add common operation policy with respond mediator
         addNewOperationPolicy();
-        Map<String, String> updatedCommonPolicyMap = restAPIPublisher.getAllCommonOperationPolicies(60, 0, null);
+        Map<String, String> updatedCommonPolicyMap = restAPIPublisher.getAllCommonOperationPolicies(70, 0, null);
```
BLasan, line 118: Same doubt here

all-in-one-apim/modules/integration/tests-common/integration-test-utils/src/main/java/org/wso2/am/integration/test/impl/RestAPIPublisherImpl.java
O-sura: With the 50 limit, the tests keeps failing because once these new policies are added the 50 limit gets exceeded resulting in consistent test failure

all-in-one-apim/modules/distribution/resources/operation_policies/specifications/apkWASMInterceptorService_v1.json
```diff
@@ -0,0 +1,85 @@
+{
+  "category": "Mediation",
+  "name": "apkWASMInterceptorService",
+  "version": "v1",
+  "displayName": "WASM Extension Service",
+  "description": "This policy allows you to call an WASM extension service for a request message",
```
Tharanidk, line 6: `an WASM` should be changed to `a WASM.`

PR #14196, opened 2026-05-12
Add a composite step to change the lifecycle status of an API or API product

VenukshiMendis:
### Description

This PR adds a composite Cucumber step to change and verify the lifecycle status of an API or API Product. It performs a given lifecycle action on the given resource and repeatedly verifies the lifecycle status until the expected state is reached or the retry limit is exceeded.


all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/PublisherCompositeSteps.java
```diff
@@ -19,18 +19,31 @@
 
 import io.cucumber.java.en.Given;
 import io.cucumber.java.en.When;
+import org.slf4j.Logger;
```
dakshina99: Do not use slf4j for logging.

VenukshiMendis: Resolved

dakshina99 (approved):

PR #14235, opened 2026-07-01
Add Integration Test Framework V2 with Concurrency Support

nimsara66:
## Overview

Introduces **integration-v2**: a capability-decomposed Cucumber + Testcontainers integration test framework for APIM that runs on a parallel-on-shared-container execution model, ports the legacy TestNG backend suite, and adds tagging/coverage tooling, framework self-verification, and CI integration to execute on every PR.

## Parallel Test Framework (Core Runtime)

- **DynamicApimContainer** — Boots APIM from the product image using dynamic (ephemeral) host ports, allowing multiple servers to run simultaneously without port collisions.
- **BlockLifecycleListener** — Boots one shared APIM container per TestNG `<test>` block, waits for a real readiness health check, provisions tenants/users for the block, and tears the container down on completion.
- **BaseBlockRunner** — Shared runner base with a boot-failure skip guard and an `@AfterClass` resource cleanup.
- **Namespaced TestContext scoping** — Provides per-block shared scope and per-runner local scope with no cross-block or cross-runner leakage under parallel execution.
- **Concurrency controls** — Capacity-bounded parallel blocks (K containers) tuned to available host/CI resources.

## Test Authoring Model

- **Actor / Identity model**
  - Provisioned actors (`admin`, `publisherUser`, `subscriberUser`) in both tenants.
  - Authentication composites mint the appropriate Publisher, DevPortal, and Admin tokens.
  - `I act as` enables mid-scenario actor switching while maintaining least-privilege by default.

- **Multi-tenancy**
  - Every capability executes against both tenants (`super` and `tenant1.com`) using `Scenario Outline` with an actor column.
  - Gateway routing uses `/t/<tenant>` paths.

- **Context helpers**
  - `${UNIQUE:...}` for collision-free resource names.
  - `{{key}}` placeholder resolution for parallel-safe tests.

- **`_setup_*` fixture pattern**
  - A runner groups a setup feature with consumer features that exchange resource IDs and payloads through runner-local scope.

- **Idempotent cleanup**
  - Per-scenario `@After("@cleanup")` hook.
  - Per-runner `@AfterClass` cleanup of registered resources.
  - Executes on failures and safely no-ops when nothing was created.

- **Gateway invocation**
  - Uses an in-network `NodeAppServer` backend.
  - Gateway-context invocation (no tenant double-prefix).
  - `...until status` retries for *N* seconds (minimum `DEPLOYMENT_WAIT_TIME`).
  - Handles transient SSL and connection failures during gateway warm-up.

- **Configuration overlays**
  - Feature-specific `deployment.toml` overlays merged via `tomlExtraOverlayPath`.
  - Supports scenarios such as:
    - Custom authentication headers
    - Application sharing
    - Token persistence
  - Enables unrelated configuration-dependent features to share the same container.

- **In-product graceful restart**
  - `ServerAdmin.restartGracefully` step.
  - Down → Up readiness verification.
  - Preserves container, ports, and database, enabling token persistence testing across restarts.

## Legacy Suite Port

- All **23 legacy product features** have been ported and reorganized by capability:
  - Publisher
  - DevPortal
  - Gateway
  - Key Manager
- Runs as an explicit per-actor, per-tenant suite.
- Improves upon the legacy suite with:
  - Comprehensive negative and role-enforcement testing (401/403 contracts)
  - Explicit two-tenant coverage
  - Least-privilege actor model
  - Multi-API-type configuration matrices (REST, SOAP, GraphQL, WebSocket)

## Tagging & Coverage Tooling

- **`capability-map.yml`**
  - Closed vocabulary of capabilities/features.
  - Source of truth for coverage.

- **`TAGGING-CONVENTIONS.md`**
  - Defines conventions for:
    - `@cap`
    - `@feat`
    - `@type`
    - `@rule`
    - `@dep`
    - `@legacy`

- **`render_coverage_tree.py`**
  - Validates tags against the capability map.
  - Fails on:
    - Unknown tags
    - `@setup` ↔ `_setup_*` mismatches
  - Generates the coverage tree.

- **`port-coverage-report.md`**
  - Legacy-to-ported comparison including:
    - Covered features
    - Intentionally dropped features
    - Newly added features
    - Remaining gaps

## Framework Self-Verification

- `verify-*.sh` and `testng-fv-*.xml`
  - Verify framework behavior including:
    - Container lifecycle
    - Scope isolation
    - Provisioning
    - Concurrency
    - End-to-end capstone scenarios
  - Prevent regressions during framework refactoring.

- **Phase 7 regression suite**
  - Verifies:
    - Context scoping
    - Multi-tenancy
    - Actor model
    - Setup handoff
    - Gateway wiring
    - Configuration overlays
    - Graceful restart
    - Taxonomy/tag validation

- Deprecated stale/legacy execution probes are guarded so regression sweeps remain green.

## CI (`.github/workflows/maven.yml`)

- **`v2-build-images`** job:
  - Builds the distribution and Docker images once.
  - Caches the Maven repository.
  - Publishes Docker images as workflow artifacts.

- **integration-v2** matrix job:
  - Downloads and loads the prebuilt Docker images.
  - Executes `mvn test` without rebuilding images.

- Both jobs execute on every PR alongside the existing legacy build jobs, which remain unchanged.

## Documentation

- **`CLAUDE.md`**
  - Test authoring guide covering:
    - Isolation
    - Cleanup
    - Actors
    - Multi-tenancy
    - Negative testing
    - Gateway invocation
    - Configuration overlays
    - Feature ordering

- **`parallel-framework-implementation-plan.md`**
  - Phased implementation roadmap and verification tracker.


nimsara66:
@coderabbitai review --dir all-in-one-apim/modules/integration-v2/tests-common/testcontainers/src/main/java/org/wso2/am/testcontainers

RakhithaRR:
@coderabbitai pause

nimsara66:
@coderabbitai review --dir all-in-one-apim/modules/integration-v2/tests-common/testcontainers/src/main/java/org/wso2/am/testcontainers

nimsara66:
@coderabbitai, Please ignore the previous review complete as it was skipped and not actually reviewed.

all-in-one-apim/modules/integration-v2/tests-common/testcontainers/src/main/java/org/wso2/am/testcontainers/APIMContainer.java
```diff
@@ -0,0 +1,154 @@
...
+        if (sharedDbUrl != null) {
+            sharedDbUrl = sharedDbUrl.replace(Constants.AMPERSAND, Constants.XML_AMPERSAND);
+        }
+
+        int offset = Constants.DEFAULT_OFFSET;
+
+        // Check if parallel execution is enabled
+        if (Boolean.parseBoolean(System.getenv(Constants.APIM_TEST_CONTAINERS_PARALLEL_ENABLED))) {
+            // Set unique offset for each container
+            offset +=  offsetCounter.getAndIncrement();
```
AnuGayan: **🔴 Blocker — read-modify-read race on the offset counter → port collisions & off-by-one DB names**

**Reason:** The offset comes from `getAndIncrement()` (pre-increment value) while the DB suffixes come from two *later* `offsetCounter.get()` reads. When multiple blocks build containers concurrently (the whole point of parallel mode), (a) the offset and DB suffix disagree by one, and the two `get()` calls can even return different values if another thread increments between them; (b) two threads can derive the **same** offset.

**Impact:** Same offset → the `addFixedExposedPort(...)` calls collide on the host and the second `container.start()` fails with a port-bind error; mismatched suffix → a container points at the wrong/nonexistent DB (`offset=500` but DB `..._1`). Parallel execution becomes flaky/broken.

**How to fix:** capture the slot once and derive everything from it:
```java
int slot = offsetCounter.getAndIncrement();
offset += slot;
apimDbUrl   = apimDbUrl.replace(Constants.APIM_DB_NAME,   Constants.APIM_DB_NAME   + "_" + slot);
sharedDbUrl = sharedDbUrl.replace(Constants.SHARED_DB_NAME, Constants.SHARED_DB_NAME + "_" + slot);
```
(Confirm DB provisioning creates the `_0` suffix for the first slot.)

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/utils/TestContext.java
```diff
@@ -0,0 +1,153 @@
...
+     * This is the single source of truth for the shared-scope key — listeners that set the scope per
+     * invocation and any onStart lifecycle wiring must both derive the key through this method.
+     */
+    public static String sharedScopeId(ITestContext ctx) {
+        return ctx.getSuite().getName() + "::" + ctx.getName();
+    }
+
+    private static final Map<String, Map<String, Object>> sharedContexts = new ConcurrentHashMap<>();
+    private static final Map<String, Map<String, Object>> localContexts = new ConcurrentHashMap<>();
+    private static final InheritableThreadLocal<ContextScope> currentScope = new InheritableThreadLocal<>();
```
AnuGayan: **🟠 Major — `InheritableThreadLocal` is not re-established on reused pool threads → cross-block scope leak**

**Reason:** `InheritableThreadLocal` copies the parent value only once, when a thread is *created* — not when a pooled TestNG worker picks up a new task. Any entry point that reaches `getScope()` without a preceding `beforeInvocation` (e.g. `onStart`/`onFinish`, `@BeforeClass`) hits the lazy branch that writes the default scope back into the ThreadLocal. Combined with the boot-failure path in `BlockLifecycleListener.onStart` not clearing scope, a stale block scope can ride a pooled worker into the next block.

**Impact:** Under parallel execution one block can read another block's shared context (wrong tokens/IDs), producing false passes/fails that the current self-verification suite structurally cannot catch (it always sets scope on the same worker that runs the body). Isolation currently holds only by happenstance, not by construction.

**How to fix:** use a plain `ThreadLocal`; make `getScope()` fail loudly (or log) instead of silently seeding the default when no scope was set; and wrap the `onStart`/`onFinish` bodies so scope is always cleared in a `finally`.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/utils/listeners/BlockLifecycleListener.java
```diff
@@ -0,0 +1,228 @@
...
+            // Start the shared backend first (idempotent singleton on the shared network) when the block opts
+            // in, so APIs deployed by gateway-invocation tests have a reachable "nodebackend" upstream.
+            if (Boolean.parseBoolean(param(context, PARAM_INIT_BACKEND))) {
+                NodeAppServer.getInstance();
+                logger.info("Block '" + label + "' ensured NodeAppServer backend is running");
+            }
+
+            DynamicApimContainer container = new DynamicApimContainer(label, resolveTomlContent(context));
+            container.withLabel("block", label);
+            container.start();
```
AnuGayan, line 112: **🟠 Major — container leaked when `start()` fails, and the failure log is misleading**

**Reason:** On the readiness-failure path the container is stopped before throwing, but if `container.start()` here throws, the reference is lost and the partially-created Docker container is never stopped. Separately, the catch logs "its classes will be skipped" (line 129) while the design (`BaseBlockRunner`) actually rethrows → the classes are **failed**, not skipped.

**Impact:** A boot failure leaks a Docker container for the JVM lifetime (relying on Ryuk to reap), and the misleading log sends anyone debugging CI in the wrong direction.

**How to fix:** hold the container in a variable visible to the `catch` and `stop()` it there (guarding against a null/already-stopped container); change the message to "…its classes will be failed".

all-in-one-apim/modules/integration-v2/tests-common/testcontainers/src/main/java/org/wso2/am/testcontainers/NodeAppServer.java
```diff
@@ -0,0 +1,69 @@
...
+
+    private static class InstanceHolder {
+        private static final NodeAppServer instance = new NodeAppServer();
+    }
+
+    public static NodeAppServer getInstance() {
+        return NodeAppServer.InstanceHolder.instance;
+    }
+
+    public void restart() {
```
AnuGayan: **🟠 Major — `restart()` on a JVM-wide singleton backend is unsynchronized and races parallel blocks**

**Reason:** `NodeAppServer` is an eager singleton shared across all blocks, and `restart()` does `isRunning() → stop() → start()` with no locking.

**Impact:** One block restarting the backend tears it down for **every** other concurrently-running block — their in-flight gateway invocations hit a stopped container, and two concurrent restarts can race on `start()`. Flaky failures unrelated to the code under test.

**How to fix:** at minimum make `restart()` (and `start`/`stop`) `synchronized`; better, reconsider whether a globally-shared backend should be restartable in the parallel lane (e.g. a per-block backend, or gate restart behind the single-container mode).

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/utils/clients/SimpleHTTPClient.java
```diff
@@ -0,0 +1,449 @@
...
+     * @return String of the response body
+     * @throws IOException throws if any error occurred
+     */
+    public static String responseEntityBodyToString(HttpResponse response) throws IOException {
+        if (response != null && response.getEntity() != null) {
+            try (InputStream inputStreamContent = response.getEntity().getContent()) {
+                return IOUtils.toString(inputStreamContent, StandardCharsets.UTF_8);
+            }
+        }
+        return null;
```
AnuGayan: **🟠 Major — empty-body responses return `null`, NPE-ing many assertion steps**

**Reason:** `responseEntityBodyToString` returns `null` when there is no entity (204/304, or a bodyless error), and `constructResponse` propagates that into `HttpResponse.getData()`.

**Impact:** Numerous steps call `response.getData().contains(...)` / `new JSONObject(response.getData())` with no null guard (e.g. `BaseSteps:503,599,642`, `PublisherBaseSteps:431,786`, `ApplicationBaseSteps:649`), so a legitimately bodyless response throws an opaque `NullPointerException` instead of producing a meaningful assertion failure.

**How to fix:** return an empty string instead of null:
```java
return "";  // instead of: return null;
```

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/utils/Utils.java
```diff
@@ -0,0 +1,928 @@
...
+    }
+
+    public static String getAllSubscriptionsURL(String baseUrl, String apiId, String applicationId, String groupId,
+                                                Integer offset, Integer limit) {
+
+        if (StringUtils.isBlank(apiId) && StringUtils.isBlank(applicationId)) {
+            throw new IllegalArgumentException("Either apiId or applicationId must be provided.");
+        }
+
+        StringBuilder urlBuilder = new StringBuilder(baseUrl + Constants.DEFAULT_DEVPORTAL + "/subscriptions");
```
AnuGayan: **🟠 Major — `getAllSubscriptionsURL` builds a double slash `//subscriptions`**

**Reason:** `Constants.DEFAULT_DEVPORTAL` already ends with a slash (`api/am/devportal/v3/`), and this line appends `"/subscriptions"` with a leading slash. Every sibling builder (e.g. line 181) uses the bare segment `"subscriptions"`.

**Impact:** The request goes to `.../devportal/v3//subscriptions`, which depending on gateway routing 404s or mis-routes — breaking `ApplicationBaseSteps.iShouldBeAbleToRetrieveSubscription`.

**How to fix:**
```java
new StringBuilder(baseUrl + Constants.DEFAULT_DEVPORTAL + "subscriptions")
```

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/utils/Utils.java
```diff
@@ -0,0 +1,928 @@
...
+    public static String getAPIDocumentContent(String baseUrl, String apiId, String documentId) {
+        return baseUrl + Constants.DEFAULT_APIM_API_DEPLOYER + "apis/" + apiId + "/documents/" + documentId + "/content";
+    }
+
+    public static String getSubscriptionBlockingURL(String baseUrl, String subscriptionID) {
+        return baseUrl + Constants.DEFAULT_APIM_API_DEPLOYER + "subscriptions/block-subscription?subscriptionId=" + subscriptionID + "&blockState=BLOCKED";
+    }
+
+    public static String getSubscriptionUnBlockingURL(String baseUrl, String subscriptionID) {
+        return baseUrl + Constants.DEFAULT_APIM_API_DEPLOYER + "subscriptions/unblock-subscription?subscriptionId=" + subscriptionID + "&blockState=BLOCKED";
```
AnuGayan: **🟠 Major — unblock-subscription URL sends `blockState=BLOCKED`**

**Reason:** `getSubscriptionUnBlockingURL` hits `unblock-subscription?...&blockState=BLOCKED` — identical to the block URL just above it. An unblock should send `UNBLOCKED`.

**Impact:** The "unblock" step is semantically wrong; the subscription can remain blocked, so any scenario asserting post-unblock behaviour silently tests the wrong state.

**How to fix:**
```java
... + "subscriptions/unblock-subscription?subscriptionId=" + subscriptionID + "&blockState=UNBLOCKED";
```

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/PublisherBaseSteps.java
```diff
@@ -0,0 +1,1663 @@
...
+     * @param scopeID Context key containing the scope ID to update
+     * @param scopePayload Context key containing the updated scope JSON payload
+     */
+    @And("I update shared scope {string} with payload {string}")
+    public void iUpdateSharedScopeWithPayload(String scopeID, String scopePayload) throws IOException {
+
+        String scopeId = Utils.resolveFromContext(scopeID).toString();
+        String jsonPayload = Utils.resolveFromContext(scopePayload).toString();
+
+        jsonPayload = jsonPayload.replace("<scopeID>", scopeID);
```
AnuGayan: **🟠 Major — `iUpdateSharedScopeWithPayload` substitutes the context *key* instead of the resolved id**

**Reason:** `scopeId` (line above) holds the resolved value, but this line replaces the `<scopeID>` placeholder with `scopeID` — the *parameter*, which is the context key name, not the value.

**Impact:** The outgoing payload contains the literal key string instead of the real scope id, so the shared-scope update sends a malformed body.

**How to fix:**
```java
jsonPayload = jsonPayload.replace("<scopeID>", scopeId);
```

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/APIInvocationSteps.java
```diff
@@ -0,0 +1,412 @@
...
+        long deadlineMillis = Math.max(timeoutSeconds * 1000L, Constants.DEPLOYMENT_WAIT_TIME);
+        long endTime = System.currentTimeMillis() + deadlineMillis;
+        do {
+            try {
+                invokeApiByContext(resolvedContext, httpMethod, accessToken, payload);
+                HttpResponse response = (HttpResponse) TestContext.get("httpResponse");
+                if (response != null && response.getResponseCode() == expectedStatus) {
+                    return;
+                }
+            } catch (Exception transientDuringWarmup) {
```
AnuGayan: **🟠 Major — the five "…until status" loops swallow all exceptions and never assert (applies to the ~5 UntilStatus methods)**

**Reason:** Each loop catches bare `Exception`, retries to the deadline, then returns **without asserting** the expected status was ever reached. It also swallows `IllegalArgumentException` from `Utils.resolveFromContext` (a missing token/payload key — a real test-authoring bug).

**Impact:** A persistent failure (bad URL/token, API never deploys) is silently swallowed; the step "passes" and the problem only maybe surfaces at a later `Then`, or not at all — a false pass. Programming errors are masked as timeouts.

**How to fix:** narrow the catch to `IOException` (the transient-connectivity case) and assert after the loop:
```java
HttpResponse last = (HttpResponse) TestContext.get("httpResponse");
Assert.assertNotNull(last, "No response captured while waiting for status " + expectedStatus);
Assert.assertEquals(last.getResponseCode(), expectedStatus,
        "API did not return " + expectedStatus + " within the deadline: " + last.getData());
```

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/utils/ResourceCleanup.java
```diff
@@ -0,0 +1,117 @@
...
+        if (TestContext.getList(Constants.CREATED_APPLICATION_IDS).isEmpty()
+                && TestContext.getList(Constants.CREATED_API_IDS).isEmpty()
+                && TestContext.getList(Constants.CREATED_OPERATION_POLICY_IDS).isEmpty()
+                && TestContext.getList(Constants.CREATED_SHARED_SCOPE_IDS).isEmpty()) {
+            return;
+        }
+        String baseUrl = baseUrlObj.toString();
+
+        try {
+            deleteResources(Constants.CREATED_APPLICATION_IDS, Identity.devportalTokenKey(Identity.defaultActor()),
```
AnuGayan: **🟠 Major — teardown always uses `defaultActor()` tokens, so resources created by other actors are not deleted**

**Reason:** Resources are registered globally (`CREATED_*_IDS`) regardless of which actor created them, but all four `deleteResources` calls resolve tokens for `Identity.defaultActor()`. In actor-switching / cross-tenant scenarios the delete is performed as the wrong principal.

**Impact:** A resource created by, say, `publisherUser@tenant1` cannot be deleted with the default actor's token (cross-tenant delete denied) → it leaks onto the shared container and 409s later runs — exactly what this class exists to prevent. (The class javadoc claims it uses "the publisher's token," which the code does not do.)

**How to fix:** record the owning actor's token key alongside each registered id and delete as that actor, rather than resolving `defaultActor()` at teardown time.

all-in-one-apim/modules/integration-v2/tests-common/testcontainers/pom.xml
```diff
@@ -0,0 +1,129 @@
...
+                        <phase>pre-integration-test</phase>
+                        <goals>
+                            <goal>exec</goal>
+                        </goals>
+                        <configuration>
+                            <executable>bash</executable>
+                            <arguments>
+                                <argument>-c</argument>
+                                <argument>
+                                    docker build ${docker.extra.hosts} https://github.com/wso2/docker-apim.git#master:dockerfiles/ubuntu/apim \
```
AnuGayan, line 115: **🟠 Major — APIM image built from a moving `wso2/docker-apim.git#master` ref**

**Reason:** The Docker build pulls the Dockerfile/context from the `master` branch of `docker-apim`, a ref that changes over time.

**Impact:** CI is non-reproducible (a green run today can turn red tomorrow with no code change here) and it's a supply-chain risk — whatever is on `master` at build time is baked into the image the tests validate.

**How to fix:** pin to a release tag or commit SHA, e.g. `...docker-apim.git#<tag-or-sha>:dockerfiles/ubuntu/apim`.

all-in-one-apim/modules/integration-v2/docs/devs/render_coverage_tree.py
```diff
@@ -0,0 +1,235 @@
...
+            indent = len(raw) - len(raw.lstrip())
+            content = raw.strip()
+            if indent == 0:
+                continue
+            if indent == 2 and content.endswith(":") and "{" not in content:
+                cur = content[:-1].strip()
+                caps[cur] = {"name": cur, "features": OrderedDict()}
+                in_feats = False
+            elif indent == 4 and content.startswith("name:"):
+                caps[cur]["name"] = content[len("name:"):].strip()
```
AnuGayan, line 50: **🟠 Major — `parse_capability_map` raises `KeyError` on a malformed map**

**Reason:** If an `indent == 4` `name:`/`features:` line appears before any capability header, `cur` is still `None`, so `caps[cur]` → `caps[None]` raises `KeyError`. The `indent == 6` branch is guarded with `cur is not None`; these two branches are not.

**Impact:** This runs as a CI gate; a slightly malformed `capability-map.yml` crashes with a raw traceback instead of a clear validation error.

**How to fix:**
```python
elif indent == 4 and cur is not None and content.startswith("name:"):
    caps[cur]["name"] = content[len("name:"):].strip()
elif indent == 4 and cur is not None and content.startswith("features:"):
    in_feats = True
```

AnuGayan (changes requested):
## Code Review — Integration Test Framework V2

Thorough review across correctness, concurrency/thread-safety, multi-instance behaviour, HTTP/REST usage, CI, poms, and a full secret scan. The architecture is sound (per-block scope namespacing, readiness gated on HTTP 200, FK-ordered cleanup, extensive self-verification). But the **headline parallel-execution feature has a genuine port-collision race**, the **thread-scope isolation is safe-by-happenstance rather than by construction**, and there are several concrete correctness bugs. All inline findings were verified against source.

### Verdict: 🔧 Request changes
Must-fix before merge: the offset/port race (`APIMContainer`) and the `InheritableThreadLocal` scope isolation. The URL/step bugs are straightforward false-pass/false-fail sources worth fixing in the same pass.

**1 blocker + 11 major findings are attached as inline comments**, each with Reason / Impact / How to fix.

---

### 🟡 Minor (not inlined)
- `APIMContainer.java:65` / `DynamicApimContainer.java:47` — NPE when DB-URL env vars are unset; fail with a clear message.
- `ContainerNetwork.java:23` — static shared `Network` never closed; all parallel containers share one network + global aliases (`nodebackend`), partially defeating per-block isolation. Document as shared, or create per block.
- `TestContext.java:94` — `addToList` mutates the `computeIfAbsent` list outside any lock; latent today but unsafe for concurrent same-scope appends (and the stress test doesn't exercise that case). Use a synchronized list + copy-under-monitor in `getList`.
- `BaseSteps.java:129` — DCR retry treats any non-200 (incl. 401/400) as retryable for the full ~300s window; break early on 4xx.
- `BaseSteps.java:397` — `putResponsePayloadInContext` NPEs on no prior response + unexplained blanket 2s sleep every call.
- `ResourceCleanup.java:110` — delete response status discarded; a 500 on delete is silently treated as success. Log non-2xx/404.
- `APIInvocationSteps.java:134` — method `switch` has no `default`; valid-but-unhandled verbs (PATCH/HEAD) silently no-op, leaving a stale response for the next assertion.
- `PublisherBaseSteps.java:282,311,1533` — `System.getenv(GATEWAY_ENVIRONMENT)` unchecked → `String.replace(null)` NPE if unset; fail fast.
- `Utils.java:374` — `getProductSearchEndpointURL` doesn't URL-encode the query (siblings do).
- `SimpleHTTPClient.java:80` — trust-all + `NoopHostnameVerifier`; fine for a self-signed test container, but scanners will flag it — suppress deliberately and confine it.
- `render_coverage_tree.py:35,79` — missing `--map` throws a raw traceback; a missing/empty `--features` dir exits 0, silently turning the lint gate into a no-op.
- Compiler drift — poms set `maven.compiler.source/target=21` and CI uses JDK 21, but the `maven-compiler-plugin` config in `integration-v2/pom.xml` and `integration-test-utils/pom.xml` pins `<source>11</source>`. Plugin config wins → those modules compile at 11. Align.
- `integration-v2/pom.xml:89` — `commons.lang3.version=3.0` is 2011-era; likely meant `3.12+`. `json-path`/`json-smart` also reference undeclared version properties.

### 🔵 Nits
- Cross-MDC mismatch: `TestNameMdcListener` uses log4j-1.x `MDC` while `DynamicApimContainer`'s consumer reads slf4j `MDC` → `testName` logs as `"default"`.
- `DynamicApimContainer.java:41` hardcodes `wso2am:4.7.0-SNAPSHOT-jdk21` default while `APIMContainer` requires the property — inconsistent, will rot.
- `PublisherBaseSteps.java:1450` empty `catch(Exception)` swallows policyId extraction failure → confusing downstream error.
- Pervasive magic literal `"httpResponse"` despite `Constants.HTTP_RESPONSE`; bearer-header map copy-pasted ~30× — extract a `bearerHeaders(token)` helper.
- `Utils.resolvePayloadPlaceholders` (~565) uses `appendReplacement` without `Matcher.quoteReplacement` (line 788 does it correctly).
- CI: `sudo echo … | sudo tee` (only `tee` needs root); consider `permissions: contents: read` on the new jobs.

---

### What's good
- Readiness gated on HTTP 200 (not just an open port); `ServerReadiness.awaitRestart` correctly handles the graceful-restart "still-serving-200" window, verified with real positive/negative controls.
- Scope storage is genuinely concurrent (nested `ConcurrentHashMap` + race-free `computeIfAbsent`), consistently namespaced by `suiteName::testName` from a single source of truth.
- `Identity` qualifies token cache keys by full `user@domain`, avoiding cross-tenant token bleed; `ResourceCleanup` orders deletes to respect FK/subscription dependencies.
- `BlockUniquenessLintListener` fails fast on namespace collisions, verified against real parsed XML.
- HTTP verbs use try-with-resources on `CloseableHttpResponse` and close entity streams — no response/connection leaks.
- No real hardcoded secrets (only standard WSO2 test defaults); the Python tool is stdlib-only (no unsafe YAML deserialization); the CI workflow has no shell-injection surface and pins its actions.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

all-in-one-apim/modules/integration-v2/tests-common/testcontainers/pom.xml
```diff
@@ -0,0 +1,129 @@
...
+                    <execution>
+                        <id>start-server</id>
+                        <phase>pre-integration-test</phase>
+                        <goals>
+                            <goal>exec</goal>
+                        </goals>
+                        <configuration>
+                            <executable>python3</executable>
+                            <async>true</async>
+                            <workingDirectory>${project.basedir}/../../../distribution/product/target</workingDirectory>
```
kavindasr, line 64: If we are building the docker image from the ZIP, why do we need this part in the workflow YAML
```
      - name: Download Docker images
        uses: actions/download-artifact@v4
        with:
          name: v2-images
          path: /tmp/v2-images
      - name: Load Docker images
        run: docker load < /tmp/v2-images/images.tar.gz
```

all-in-one-apim/modules/integration-v2/tests-common/testcontainers/src/main/java/org/wso2/am/testcontainers/APIMContainer.java
nimsara66: "APIMContainer.java" has no real use in the updated concurrent approach. Therefore will remove this from this PR.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/utils/TestContext.java
nimsara66: The `InheritableThreadLocal` semantics you cite are correct in isolation, but the scope is set fresh for every invocation:

`BlockScopeListener.beforeInvocation` binds the scope on the same worker thread that executes each method (both test and configuration methods—it also covers `@BeforeClass` and `@AfterClass`), and `afterInvocation` clears it. As a result, a stale value on a reused pool thread is always overwritten before any method body reads it. Likewise, `onStart` and `onFinish` set the scope explicitly before any access, so they never reach the lazy-default branch.

This behavior is regression-guarded by `BlockScopeThreadVisibilityVerificationTest` (FV-2.2), which runs parallel blocks with a parallel data provider and a poisoned global scope, asserting that no invocation reads either the default or a sibling's scope—the exact failure mode described. This was also confirmed empirically: 6 invocations executed across 2 reused pool threads spanning 3 blocks, with zero scope leaks.

By design inheritance was never the isolation mechanism. As a result, FV-2.2 also passes with a plain `ThreadLocal`. Therefore adopted your suggestion and switched to a plain `ThreadLocal` because it is simpler and fails closed to the global default rather than risking a stale inherited snapshot.

Applied:
- `TestContext.currentScope`: changed from `InheritableThreadLocal` to a plain `ThreadLocal`, with a comment explaining that the scope is strictly per invocation.
- `BlockLifecycleListener.onStart`: now clears the scope in a `finally` block (mirroring `onFinish`).

Not changed:
- `getScope()` still seeds the default when no scope is bound. The global default is a legitimately used scope (e.g., FV-2.2's poison trap), so failing loudly there would break valid callers.

Verified:
- All scope-related FV suites (`2.1`, `2.2`, `2.3`, `7.1`, and `C.2`) pass.
- The full integration suite passes **168/168** tests at **TP=2 / CP=2** (concurrent multi-block load), with no scope-related failures.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/utils/listeners/BlockLifecycleListener.java
nimsara66: Verified and fixed.

On the boot-failure path, the container reference was scoped inside the `try` block, so the `catch` block could not stop a container that had already started before the failure occurred (e.g., during `start()`, URL retrieval, or a readiness check). As a result, the container was leaked until Ryuk eventually cleaned it up.

This has been fixed by hoisting the container reference outside the `try` block so it is visible to the `catch` block. The cleanup is guarded against `null`, and the `stop()` call is wrapped so that any cleanup failure cannot mask the original boot exception.

The log message was also corrected from _"its classes will be skipped"_ to _"its classes will be reported as failed"_ to accurately reflect the behaviour of `BaseBlockRunner`, which re-throws the boot exception, causing the build to fail rather than skipping the affected classes.

Verified running against the boot-failure FV suites (`4.8` and `4.9`), which include assertions that no container with the corresponding block label remains after a boot failure.

all-in-one-apim/modules/integration-v2/tests-common/testcontainers/src/main/java/org/wso2/am/testcontainers/NodeAppServer.java
nimsara66: Removed the `restart()` functionality from the node server, as it is not currently used. 

With the concurrent execution model, if restarting a node server becomes a requirement in the future, the correct approach would be to use block-level node servers. This is now feasible because this effort reduced the original node server image size from **1.3 GB** to **less than 200 MB**.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/utils/Utils.java
nimsara66: Verified and fixed.

The double slash is indeed present: `DEFAULT_DEVPORTAL` ends with `/`, and `getAllSubscriptionsURL` appended `"/subscriptions"`, making it inconsistent with its sibling methods (`getCreateSubscriptionURL` and `getSubscriptionURL`), which append the segment without a leading slash. This has been corrected to `+ "subscriptions"`.

The reported impact, however, does not hold. This is a Dev Portal REST call to the Tomcat/CXF servlet at `api/am/devportal/v3`, not a gateway-routed request, and that stack normalizes `//` in the request path. This was also verified empirically: `iShouldBeAbleToRetrieveSubscription`—which strictly throws `"No subscription found"` unless the response contains a non-empty list—executes through `subscribe.feature` and passes as part of the full integration suite (**168/168** passing). In other words, `//subscriptions` was already resolving successfully with an HTTP 200 response.

The fix is still worthwhile for consistency and defensiveness, as stricter proxies or routers may not normalize `//` in the same way. The change has therefore been applied, but this was not a functional issue in practice.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/utils/Utils.java
nimsara66: Fixed—though the diagnosis needs correcting on two points, both confirmed against the shipped specification.

The copied `blockState=BLOCKED` parameter on the unblock URL is indeed present. However, it is not responsible for causing a subscription to remain blocked. The `unblock-subscription` operation in `publisher-api.yaml` (from `api#am#publisher.war` in the built distribution) declares only two parameters: `subscriptionId` and `If-Match`. It does not define a `blockState` parameter. As a result, the extra query parameter is ignored by CXF. This is also consistent with our `security_enforcement` scenario (`block → 401 → unblock → 200`), which passes for both tenants in the full integration suite (**168/168** passing), confirming that the unblock operation correctly restores access.

Additionally, `UNBLOCKED` is not a valid value. The `blockState` parameter exists only for the `block-subscription` operation, where its valid enum values are `BLOCKED` and `PROD_ONLY_BLOCKED` (the specification's own example uses `PROD_ONLY_BLOCKED`). Therefore, the correct fix is to remove the parameter entirely rather than change its value. The unblock URL now sends only the `subscriptionId`, matching the specification's `curl` example.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/PublisherBaseSteps.java
nimsara66: Removed the step definition as it is dead code.

all-in-one-apim/modules/integration-v2/docs/devs/render_coverage_tree.py
nimsara66: Reproduced the `KeyError(None)`: in the `indent == 4` / `name:` branch, `caps[cur]["name"]` is accessed while `cur` is still `None` when the entry appears before any capability header. Unlike the `indent == 6` branch, this case is unguarded.

Minor precision: only the `name:` branch actually raises; the `features:` branch is also unguarded but does not dereference `caps[cur]`—it only sets `in_feats`.

It already fails loudly (non-zero exit code with a traceback), so there is no false pass—only an unclear error message, and it occurs only under an unusual, manually malformed input.

On the fix: instead of adding a guard-and-skip (which would allow this script—whose purpose is to reject malformed maps—to silently ignore broken structure), I changed it to fail explicitly using `sys.exit(...)` in both `indent == 4` branches. Verified behaviour: the valid capability map still lints cleanly (`invalid: 0`), and a `name:`-before-header map now exits non-zero with a clear error message: `capability-map malformed: 'name:' at indent 4 before any capability header` instead of raising a `KeyError`.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/utils/ResourceCleanup.java
nimsara66: Verified — this was a bug, not just a latent issue. 

Added teardown observability (checking the delete response, not just catching exceptions), and the current design leaked six resources in a single run: two applications were deleted using the wrong actor’s token, resulting in 403 Forbidden, and four APIs failed with 500 errors — all silently. The best-effort teardown combined with `${UNIQUE}` names hid the failures, so the suite still reported green.

Note that `defaultActor()` is actually the current acting actor, not a fixed default — but the effect is exactly what you described when the actor at teardown differs from the creator.

This was fixed by recording the creating actor against each registered ID (`ResourceCleanup.register → OwnedResource`) and deleting each resource using its original actor. 

On re-run: zero failed deletes, 168/168 green. This also removes the previous requirement to “remember to switch back to the owner before @cleanup” during authoring.

| Signal                               | Before (acting-actor teardown) | After (owner-aware teardown) |
|--------------------------------------|--------------------------------|-------------------------------|
| Resource NOT deleted (403/500 leaks) | 6 (2 apps: 403, 4 APIs: 500)   | 0                             |
| No token for the owner (new guard)   | —                              | 0                             |
| Transient cleanup failures           | 0                              | 0                             |
| 404 (already deleted)                | 5                              | 5                             |

all-in-one-apim/modules/integration-v2/tests-common/testcontainers/pom.xml
nimsara66: The `integration-v2` (test) job does **not** build the Docker image—the separate `v2-build-images` job builds it once. The download/load step is simply how that prebuilt image is transferred from the build job to the test job.

The reason this is necessary is that GitHub Actions jobs run on independent runners, each with its own Docker daemon. An image built (from the ZIP) in `v2-build-images` exists only in that runner's Docker daemon; it is not visible to the `integration-v2` job's daemon. There is no shared local Docker registry between jobs. To bridge this gap, the build job runs `docker save … | gzip` and uploads the resulting tarball as the `v2-images` artifact. The test job then performs `download-artifact` followed by `docker load` to materialize that exact image in its own Docker daemon before running `mvn test`.

The workflow is split this way because building the distribution and Docker image is the expensive step. By doing it once and reusing the result, we avoid rebuilding the image in every test job. The test job runs `mvn test` (not `mvn install`), which is intentionally designed to reuse the prebuilt image rather than rebuild it. If the image were not loaded first, Testcontainers would have no `wso2am:…` image to start, and the test run would fail. Loading the saved artifact also guarantees that every test job runs against the exact byte-identical image produced by the build job.

In short, building from the ZIP happens only in `v2-build-images`; the download/load step is the hand-off mechanism that makes the image available to the test job. Without it, the test job would have no image to run against.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/APIInvocationSteps.java
nimsara66: Fixed — slightly beyond the suggested fix, because the shared `httpResponse` context key introduced a second false-pass scenario that the original snippet would not have caught.

Verified the issue: each retry loop caught a bare `Exception` and returned without asserting, so a persistent failure only surfaced (if at all) in a subsequent `Then` step. Additionally, a missing token or payload key (an `IllegalArgumentException` from `Utils.resolveFromContext`) was swallowed and masked as a timeout.

Identified a more subtle issue: on a persistent exception, the loop never updates `httpResponse`, so it retains a stale value from a previous step. As a result, even a post-loop `assertEquals` against the shared context key could incorrectly pass using leftover data.

**What changed (`APIInvocationSteps`):**
- Funnelled all invocations through a single `execute(...)` primitive that clears `httpResponse` first, performs the invocation, sets the context only when a real response is received, and returns the response. This also eliminated approximately 15 duplicated `set("httpResponse", …)` calls.
- Updated all five retry loops to catch only `IOException`, allowing transient warm-up failures to be retried while causing invalid token/key lookups to fail fast instead of timing out.
- Each retry loop now retains the last successful response in a local variable and performs a shared post-loop assertion:
  - `assertNotNull(last, …)` to detect when every attempt failed.
  - `assertEquals(last.getResponseCode(), expected, …)` to verify the final response status.

As a result, each step now fails independently and cannot be fooled by a stale shared response.

AnuGayan (approved):

PR #14237, opened 2026-07-01
Make MCP Server tool-list assertions order-independent

kavindasr:
## Purpose

The MCP Server integration tests (`MCPServerTestCase`) asserted `tools/list` responses and tool schemas by comparing against exact JSON strings. Since the order of tools in the response — and the order of keys within a schema — is not guaranteed, these assertions were flaky.

## Approach

- Assert on the **presence of expected tools by name** (parsing the response with Jackson) instead of matching the full JSON string, in:
  - `testMCPServerSubscribeAndInvokeForDirectBackendSubtype` (`get_pets`, `get_pets_by_petId`)
  - `testToolsForExistingApiSubtype` (`delete_oldpets`, `get_pets`)
  - `testToolsForProxySubtype` (`echo`, `orderPizza`)
- Compare the `delete_oldpets` schema **structurally** via a new `compareJson` helper rather than by exact string match.
- Reference the shared `TARGET_*` constants for tool names in the WireMock `tools/list` envelope.

## Notes

Test-only change; no product code is affected.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nimsara66 (changes requested):
@kavindasr, Can you implement this in V2 framework

kavindasr:
@nimsara66 
This change fixes an existing bug. Are you asking me to remove this test case?

AnuGayan (approved):

AnuGayan:
@nimsara66 Since this is an existing test case, is this problem fixed in the migrated tests?

kavindasr:
Fixed by: https://github.com/wso2/product-apim/pull/14233

PR #14240, opened 2026-07-06
Implement code coverage listener and coverage skill

nimsara66:
This pull request introduces a comprehensive test authoring skill for maximizing test coverage and regression protection in the APIM project. It provides detailed operating principles, a multi-phase workflow, and supporting reference documents to guide the creation, placement, and verification of both unit and integration tests. Additionally, it updates the CI workflow to collect and upload integration test coverage reports.

**Test Coverage Authoring Skill and Workflow**

- Introduces `.claude/skills/test-coverage/SKILL.md`, a detailed guide for a test authoring skill that covers intake, change analysis, gap analysis, test-plan synthesis, approval, implementation, verification, and wrap-up phases. It emphasizes maximizing coverage, minimizing duplication, and enforcing lead approval for costly or vocabulary-extending changes.
- Adds supporting reference documents for each phase:
  - [`change-analysis.md`](diffhunk://#diff-99934737ac1b52cc875b3eacf1791e2287e36dd5497e70e7962e27f617968c2aR1-R36): How to break down diffs into test-worthy semantic units and classify them.
  - [`gap-analysis.md`](diffhunk://#diff-28ec1fc393483ea5ef063c6c01a93f8cdf82cf67609399fc00fdfc440e5c8405R1-R35): Methodology for identifying coverage gaps in unit and integration tests.
  - [`duplication-check.md`](diffhunk://#diff-4018e7a1922360b6e0260047c694878f7af33b16c244ea86904e889183bddf0aR1-R30): Checklist and procedures to avoid duplicate steps, features, and blocks in integration tests.
  - [`plan-template.md`](diffhunk://#diff-99904945a142d02642c4de300f967b80a341722398b4ef05522ee076efc8bd04R1-R41): Markdown template for presenting test plans, with clear selection and approval gates.
  - [`repo-config.md`](diffhunk://#diff-8ae5913414fdc3229bf27a34223334dee3b23ab74b601462b803321f701c66a6R1-R25): Format and rules for storing local repo paths to coordinate multi-repo analysis and testing.

**Continuous Integration Enhancements**

- Updates the Maven workflow to enable coverage collection for integration tests by adding `-Dapim.coverage=true` to test runs.
- Archives the generated coverage output and uploads integration test coverage reports to Codecov for visibility and tracking.

These changes provide a robust framework for systematic, high-value test coverage improvements and ensure that integration test coverage is continuously measured and reported.


all-in-one-apim/modules/integration-v2/docs/devs/v2-coverage-architecture.md
```diff
@@ -0,0 +1,174 @@
...
+    --xml coverage/output/jacoco-it.xml --html coverage/output/html
+open coverage/output/html/index.html      # total % also printed to stdout
+```
+Confirms: agent attached, tcpserver dump over mapped port worked, classfiles/sources map correctly.
+
+### 6.2 Level 2 — validate the Codecov config offline (nothing sent)
+```bash
+curl -s --data-binary @codecov.yml https://codecov.io/validate      # validate yaml
+codecovcli do-upload --dry-run --flag integration-v2_tests --file coverage/output/txt/jacoco-it.xml
+```
```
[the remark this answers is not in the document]
nimsara66: @coderabbitai,
Fixed — all jacoco-it.xml references now use coverage/output/txt/jacoco-it.xml (the real report path); note the dry-run was already correct, so the actual inconsistency was in the diagram/Level-1 generation commands

all-in-one-apim/modules/integration-v2/tests-common/testcontainers/src/main/java/org/wso2/am/testcontainers/DynamicApimContainer.java
```diff
@@ -109,6 +112,39 @@ public DynamicApimContainer(String containerLabel, String deploymentTomlContent)
...
+    public DynamicApimContainer withCoverage() {
+        try {
+            File agentJar = JacocoCoverage.extractAgentJar();
+            withCopyToContainer(MountableFile.forHostPath(agentJar.getAbsolutePath()),
+                    JacocoCoverage.CONTAINER_AGENT_PATH);
+        } catch (IOException e) {
+            throw new RuntimeException("Failed to stage JaCoCo agent jar", e);
+        }
+        // JAVA_TOOL_OPTIONS is honoured by the JVM itself, so no need to touch the WSO2 startup script.
+        withEnv("JAVA_TOOL_OPTIONS", JacocoCoverage.containerAgentVmArg());
```
AnuGayan: 🔴 **Blocker — `JAVA_TOOL_OPTIONS` attaches the agent to every JVM in the container; the tcpserver port collision aborts JVMs and fails block boots**

**Impact:** `JAVA_TOOL_OPTIONS` is honoured by *every* JVM launched in the container, not just the server. The first JVM binds JaCoCo's tcpserver port 6300; every subsequent JVM gets `java.net.BindException: Address already in use` in the agent's `premain` and aborts with `FATAL ERROR in native method: processing of -javaagent failed`. When the aborted JVM is one the boot path needs, the block never becomes ready. This PR's own CI run (28836820533) shows exactly that: `Block 'devportal' boot/readiness failed`, `APIM block boot failed` across 6 runners → **7 failed / 24 skipped**. Since the workflow now passes `-Dapim.coverage=true` unconditionally, every future PR's integration-v2 job inherits this. Subtler: even in blocks that survive, whichever JVM wins the 6300 race is the one that gets dumped — if a short-lived helper JVM wins, the dump silently contains near-zero server coverage.

**How to verify:** open the integration-v2 job log of this PR's CI run (or run `mvn test -pl tests-integration/cucumber-tests -Dsurefire.suite.xml=testng-v2.xml -Dapim.coverage=true` locally) and search container output for `Picked up JAVA_TOOL_OPTIONS` followed by `BindException: Address already in use` and `FATAL ERROR in native method: processing of -javaagent failed`.

**Solution:** scope the agent to the server JVM only instead of the process-global env var — e.g. pass `-javaagent:...` through the same server-only channel the debug flag already uses (`withCommand(...)` around line 99), or a startup-script/env hook that only `wso2server.sh` reads. That both fixes the boot failures and guarantees the dump always targets the server JVM.

.github/workflows/maven.yml
```diff
@@ -153,6 +153,16 @@ jobs:
...
           retention-days: 1
           if-no-files-found: error
+      # Coverage report rendering needs the product class files, which JacocoCoverage extracts from the built
+      # distribution zip. The integration-v2 job restores only the m2 cache + Docker images, so share the zip
+      # here too (same pattern as the images) — without it the report can't render (`Distribution zip not found`).
+      - name: Upload APIM distribution zip (for coverage report)
+        uses: actions/upload-artifact@v4
+        with:
+          name: v2-dist-zip
+          path: all-in-one-apim/modules/distribution/product/target/wso2am-4.7.0-SNAPSHOT.zip
```
AnuGayan: 🟡 **Minor — hardcoded distribution zip name will break on the next version bump**

**Impact:** this artifact path hardcodes `wso2am-4.7.0-SNAPSHOT.zip`, while the consumer (`CoverageSupport.distributionZip()`) derives the name from `-Dapim.server.name`. On a version bump, this step fails (`if-no-files-found: error` — at least loudly), and if only one side is updated the coverage report silently stops rendering again, which is exactly the failure mode this commit just fixed.

**How to verify:** bump the product version and run the workflow — the upload step errors out, or (if the glob side is fixed but not the other) the integration-v2 job logs `Distribution zip not found`.

**Solution:** use a glob (`all-in-one-apim/modules/distribution/product/target/wso2am-*.zip`) or derive the file name from the same version property in both the upload step and the consuming code path.

.github/workflows/maven.yml
```diff
@@ -205,4 +224,20 @@ jobs:
             all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/target/surefire-reports/**
             all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/target/cucumber-report/**
             all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/target/logs/**
+            all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/target/coverage/output/**
           if-no-files-found: warn
+      - name: Upload v2 integration coverage to Codecov
+        if: always()
+        uses: codecov/codecov-action@v5
+        with:
+          files: all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/target/coverage/output/txt/jacoco-it.xml
```
AnuGayan, line 257: 🟠 **Major — Codecov upload fails silently, needs a token for non-fork runs, and the planned `codecov.yml` is missing**

**Impact:** `codecov/codecov-action@v5` defaults to `fail_ci_if_error: false`, so an upload that finds no report still shows a green step. That already happened in this PR's CI run 28836820533: the step logged `Some files were not found --- jacoco-it.xml`, `Found 0 coverage files to report`, `Error: No coverage reports found` and `Failed to run upload-coverage` — yet passed. Nobody will notice if coverage regresses to zero-upload again. Separately: v5 tokenless upload only works for **fork** PRs (`CC_FORK: true` in the log); same-repo PRs and `workflow_dispatch` runs will need `CODECOV_TOKEN`. And the design doc (v2-coverage-architecture.md §6.2) plans a `codecov.yml` with `carryforward: true` + `informational: true` for this flag — the repo has none, so once uploads work the new `integration-v2_tests` flag gets Codecov's default status-check behavior.

**How to verify:** expand the "Upload v2 integration coverage to Codecov" step in run 28836820533 — the upload failure is logged but the step is green; check the Codecov project for the absent `integration-v2_tests` flag.

**Solution:** add `fail_ci_if_error: true` (or at minimum surface a warning), pass `token: ${{ secrets.CODECOV_TOKEN }}`, and commit the `codecov.yml` proposed in the design doc.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/utils/listeners/CoverageAggregationListener.java
```diff
@@ -0,0 +1,90 @@
...
+ *
+ * <p>No-op when coverage is off, so it can stay registered in any suite without affecting normal runs.
+ * All failures are logged and swallowed — coverage reporting must never turn a green suite red.
+ */
+public class CoverageAggregationListener implements ISuiteListener {
+
+    private static final Log logger = LogFactory.getLog(CoverageAggregationListener.class);
+
+    @Override
+    public void onStart(ISuite suite) {
```
AnuGayan, line 45: 🟡 **Minor — stale `.exec` files from previous runs are silently merged into the report**

**Impact:** `onFinish` merges every `*.exec` found in `target/coverage/exec/`, but nothing purges that directory at suite start. A local rerun without `mvn clean` merges dumps from a previous (possibly different) build, inflating the coverage numbers. The test-coverage skill doc even tells the developer to "purge `target/coverage/exec/` first" as a manual step — that discipline belongs in code.

**How to verify:** run the suite twice with `-Dapim.coverage=true` without cleaning; the second run's `Aggregating coverage from N block exec file(s)` log line shows double the exec files.

**Solution:** in `onStart(ISuite)`, when `CoverageSupport.enabled()`, delete the exec directory before the suite runs (same clean-slate approach `extractApimgtClassfiles` already takes for the classfiles dir via `deleteRecursively`).

all-in-one-apim/modules/integration-v2/tests-common/testcontainers/src/main/java/org/wso2/am/testcontainers/JacocoCoverage.java
```diff
@@ -0,0 +1,342 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(JacocoCoverage.class);
+
+    /** TCP port the in-container agent listens on (canonical; Docker maps it to an ephemeral host port). */
+    public static final int TCP_PORT = 6300;
+    /** Where the agent jar is copied inside the container. */
+    public static final String CONTAINER_AGENT_PATH = "/opt/cov/jacocoagent.jar";
+
+    /** Default scoping — instrument APIM code only, skip generated/stub/test classes. */
+    public static final String DEFAULT_INCLUDES = "org.wso2.carbon.apimgt.*";
+    public static final String DEFAULT_EXCLUDES = "*.stub.*:*.dto.*:*.gen.*:*Test";
```
AnuGayan: 🟡 **Minor — agent excludes vs report denominator mismatch systematically understates coverage**

**Impact:** the agent skips instrumenting `*.stub.*:*.dto.*:*.gen.*`, but `extractApimgtClassfiles` puts all of those classes into the report denominator. Excluded-but-executed classes are therefore reported as 0% forever, understating per-class and overall coverage. (The doc's own rule of thumb — "if the classfiles set is right, the % is correct even if the agent instrumented slightly more" — only holds when the agent instruments a *superset*; here it instruments a *subset*.) Also, the architecture doc §4.1 lists `*.thrift.*` in the excludes but the code omits it.

**How to verify:** run any DTO-heavy REST flow with coverage on and open the HTML report — `.dto.` packages show 0% despite being exercised.

**Solution:** apply the same exclude patterns when staging/analyzing classfiles so instrumented set == denominator (or drop the agent-level dto/stub/gen excludes), and align the excludes string with the doc.

all-in-one-apim/modules/integration-v2/docs/devs/v2-coverage-architecture.md
```diff
@@ -0,0 +1,174 @@
+# v2 integration-test coverage — architecture & local verification (all-in-one)
+
+**Status:** Implemented — opt-in `-Dapim.coverage=true` drives agent-at-boot → per-block dump → suite-end merge+report (verified locally end-to-end). The CI wiring is committed too: the `integration-v2` job runs the suite with coverage and has a Codecov upload step (flag `integration-v2_tests`). **One CI gap remains** before it is end-to-end in CI: that job doesn't yet have the built distribution zip on disk, so the report can't render there (§8).
```
AnuGayan: 🔵 **Nit — status line is stale relative to the latest commit**

**Impact:** this line says "One CI gap remains … that job doesn't yet have the built distribution zip on disk", but the "Share distribution zip with integration-v2 CI jobs" commit in this same PR adds exactly that wiring (`v2-dist-zip` upload/download in maven.yml). The doc now understates what's implemented and will mislead the next reader.

**How to verify:** compare this line with the `v2-dist-zip` steps in `.github/workflows/maven.yml`.

**Solution:** update the Status line (and the matching note in §8) to say the zip is now shared with the integration-v2 job.

all-in-one-apim/modules/integration-v2/tests-common/testcontainers/src/main/java/org/wso2/am/testcontainers/JacocoCoverage.java
```diff
@@ -0,0 +1,342 @@
...
+                    try (java.util.zip.ZipInputStream war =
+                                 new java.util.zip.ZipInputStream(zip.getInputStream(e))) {
+                        ZipEntry we;
+                        while ((we = war.getNextEntry()) != null) {
+                            String wn = we.getName();
+                            if (we.isDirectory()) {
+                                continue;
+                            }
+                            if (wn.startsWith("WEB-INF/classes/org/wso2/carbon/apimgt/") && wn.endsWith(".class")) {
+                                File out = new File(webappClasses, wn.substring("WEB-INF/classes/".length()));
```
AnuGayan: 🟡 **Minor — Zip-Slip hardening: WAR entry names are not checked for `..` segments**

**Impact:** the output path is built from the WAR entry name without rejecting `..` segments — an entry like `WEB-INF/classes/org/wso2/carbon/apimgt/../../../../x.class` passes the `startsWith`/`endsWith` checks and escapes the `webapp-classes` directory (classic Zip-Slip). The input is a locally built distribution zip, so real exploitability is low, but security scanners (and CodeRabbit) will flag it.

**How to verify:** craft a test WAR with a `..` entry under `WEB-INF/classes/org/wso2/carbon/apimgt/` and observe the extracted file landing outside `destDir`.

**Solution:** after resolving `out`, require `out.getCanonicalPath().startsWith(webappClasses.getCanonicalPath() + File.separator)` (and the same for the lib-jar branch) before copying; skip/log entries that fail.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/utils/CoverageSupport.java
```diff
@@ -0,0 +1,85 @@
...
+        return root(moduleDir).resolve(Paths.get("output", "txt", "jacoco-it.xml").toString()).toFile();
+    }
+
+    public static File outputHtml(String moduleDir) {
+        return root(moduleDir).resolve(Paths.get("output", "html").toString()).toFile();
+    }
+
+    /** The built all-in-one distribution zip (source of the report's class files). */
+    public static File distributionZip(String moduleDir) {
+        String serverName = System.getProperty("apim.server.name"); // e.g. wso2am-4.7.0-SNAPSHOT
```
AnuGayan, line 113: 🟡 **Minor — missing `apim.server.name` produces a confusing `null.zip` path instead of a clear error**

**Impact:** if `apim.server.name` isn't set (e.g. a runner invoked outside the Maven config that forwards it), `serverName` is `null` and this resolves `…/target/null.zip`. The aggregation listener then logs `Distribution zip not found (…/null.zip)` — which hides the actual cause (unset property) behind a bizarre-looking path.

**How to verify:** run the coverage smoke suite without `-Dapim.server.name`; the warn log shows the `null.zip` path.

**Solution:** fail fast with an explicit message, e.g. `throw new IllegalStateException("apim.server.name not set - required to locate the distribution zip for coverage reporting")`, or log the missing property by name before returning.

AnuGayan (commented):
## Review — coverage listener & CI wiring

The architecture (agent-at-boot → per-block tcpserver dump → suite-end merge/report) and the docs are well thought out, and the latest commits already fix the missing-distribution-zip gap in CI. However, **the agent injection via `JAVA_TOOL_OPTIONS` demonstrably breaks test-block boots** (see inline comment — verified in this PR's own CI run 28836820533: 7 failed / 24 skipped with `APIM block boot failed`), and the Codecov upload fails silently. Overall assessment: 🔧 needs changes before the always-on `-Dapim.coverage=true` in CI is safe.

Inline comments are formatted as **Impact / How to verify / Solution**. Findings: 1 blocker, 1 major, 5 minor, 1 doc nit.

Smaller nits (no inline comment): ignored `mkdirs()` return values in `JacocoCoverage.dump`/`report`; inline fully-qualified names (`java.util.Set`, `java.util.zip.ZipInputStream`, `java.util.stream.Stream`) instead of imports; the Codecov XML lives under `output/txt/` (an XML in a `txt/` dir — if this mirrors a legacy convention, a comment would help); `extractApimgtClassfiles` is pure zip logic that would be easy to unit-test with a synthetic zip.

What's good: dump-before-`stop()` ordering is correct; coverage failures are logged-and-swallowed so they can't turn a green suite red; `addExposedPort` preserves the canonical port set; clean-slate classfile re-extraction with first-wins WAR dedupe; per-root analyzer resilience; unusually thorough design docs.

all-in-one-apim/modules/integration-v2/tests-common/testcontainers/src/main/java/org/wso2/am/testcontainers/DynamicApimContainer.java
nimsara66: Confirmed and fixed. 

Root cause is exactly as described — JAVA_TOOL_OPTIONS is inherited by every JVM, so the backgrounded diagnostics-tool + helper JVMs race the Carbon JVM for tcpserver port 6300 → BindException → FATAL …  -javaagent failed.

Note the suggested withCommand(-javaagent…) doesn't work either: api-manager.sh places command args after org.wso2.carbon.bootstrap.Bootstrap, so they're program args, not JVM options. 

Fixed by injecting the agent via JVM_MEM_OPTS, which api-manager.sh applies to the main Carbon JVM only (line 326); diagnostics.sh overwrites JVM_MEM_OPTS with its own value so the diagnostics JVM never gets the agent. 

Verified on a full TP=2 run: 0  BindException, 0 FATAL, all 6 blocks boot, 168 tests pass, 6/6

.github/workflows/maven.yml
nimsara66: Fixed — the upload path derives the name from the pom's apim.server.name (the same property CoverageSupport.distributionZip() consumes), so both producer and consumer track a version bump from one source and there's no hardcoded 4.7.0-SNAPSHOT left.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/utils/listeners/CoverageAggregationListener.java
nimsara66: Fixed — onStart(ISuite) now purges the exec dir via JacocoCoverage.deleteRecursively when coverage is enabled (same clean-slate approach extractApimgtClassfiles already uses). 

Verified on the full run: "Aggregating coverage from 6 block exec file(s)" — no stale inflation. The manual "purge first" step was removed from the skill doc.

.github/workflows/maven.yml
nimsara66: Addressed in parts: 

(a) Silent upload — added a pre-upload "Verify coverage report was generated" step that fails the job if jacoco-it.xml is missing/empty/malformed, so a zero-upload can't stay green; the original zero-report cause (zip not on disk in the job) is also fixed. 

(b) Token — intentionally omitted: this lane runs on fork PRs, which can't read  repo secrets, and codecov-action v5 tokenless works for exactly those fork PRs. 

(c) codecov.yml — deferred; the integration-v2_tests flag uses Codecov defaults for now, with the carryforward/informational shape kept in the design doc for when we enable per-flag tuning.

all-in-one-apim/modules/integration-v2/tests-common/testcontainers/src/main/java/org/wso2/am/testcontainers/JacocoCoverage.java
nimsara66: Fixed (Option A) — report() now applies the same DEFAULT_EXCLUDES to the analyzed denominator via analyzeFiltered, so stub/dto/gen/thrift classes no longer sit at a forced 0%. Added *.thrift.* to align the code with the doc. 

Verified: denominator drops 3005 → 1994 classes and the figure rises to the accurate ~25.2% (previously understated).

all-in-one-apim/modules/integration-v2/docs/devs/v2-coverage-architecture.md
nimsara66: Fixed — updated the Status line and the matching note to reflect that the product class files are now shared with the integration-v2 job.

all-in-one-apim/modules/integration-v2/tests-common/testcontainers/src/main/java/org/wso2/am/testcontainers/JacocoCoverage.java
nimsara66: Fixed — added an isWithin(out, webappClasses) canonical-path containment check on the webapp-classes branch and a symmetric guard on the lib branch; entries resolving outside the extraction dir are skipped and logged. 

Verified: 0  zip-slip skips on the real distribution (no legitimate entry affected).

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/utils/CoverageSupport.java
nimsara66: Fixed — distributionZip() now throws IllegalStateException with an explicit message when apim.server.name is unset, instead of resolving …/null.zip and surfacing a misleading "not found" later.

AnuGayan (approved):

PR #14264, opened 2026-07-19
Port Group 3 backend tests into the integration-v2 Cucumber framework

nimsara66:
This pull request significantly expands and updates the API integration test coverage documentation, reflecting new scenarios, features, and improved organization in the capability map and coverage tree. The main focus is on increased scenario coverage, new features around API visibility and versioning, and more granular test cases for DevPortal and Publisher functionality.

**Key changes include:**

### Coverage Expansion & Organization

* Increased the total number of product scenarios from 299 to 399, with new and more detailed test cases added across Publisher and DevPortal features.
* Expanded scenario counts for major functional areas, such as Publisher API lifecycle, API config, versioning, definitions, and DevPortal discovery, applications, and visibility. [[1]](diffhunk://#diff-39b77fe3d81db566bad1b67b2d8849fbac066f56e590f2694542d003d1524a07L5-R33) [[2]](diffhunk://#diff-39b77fe3d81db566bad1b67b2d8849fbac066f56e590f2694542d003d1524a07L109-R178)

### API Visibility & Access Control

* Added a new `visibility` feature to the capability map for store-side (DevPortal) role-based API visibility, ensuring restricted APIs and their documentation are only accessible to authorized users.
* Introduced and documented new test cases for both publisher-side and store-side API visibility, covering scenarios like restricted/public API discovery, access control, and tag visibility for unauthorized users. [[1]](diffhunk://#diff-39b77fe3d81db566bad1b67b2d8849fbac066f56e590f2694542d003d1524a07L97-R120) [[2]](diffhunk://#diff-39b77fe3d81db566bad1b67b2d8849fbac066f56e590f2694542d003d1524a07L109-R178)

### Publisher Feature Enhancements

* Added new and more granular test cases for API lifecycle (e.g., version copying, provider changes), API configuration (e.g., endpoint certificates, null fields, metadata), operation policies (e.g., secret attributes), and API definitions (e.g., resource modification, cross-tenant import). [[1]](diffhunk://#diff-39b77fe3d81db566bad1b67b2d8849fbac066f56e590f2694542d003d1524a07R45-R63) [[2]](diffhunk://#diff-39b77fe3d81db566bad1b67b2d8849fbac066f56e590f2694542d003d1524a07L58-R103)
* Expanded SOAP API design and service catalog test coverage, including WSDL import/export and service catalog pagination/import scenarios.

### DevPortal Feature Enhancements

* Expanded DevPortal test cases for discovery (e.g., content search, tag visibility, search syntax), applications (e.g., duplicate name rejection, owner isolation), and store-side visibility.
* Added new documentation for application attributes and improved application CRUD scenarios.

### Documentation and Test Structure Improvements

* Updated the coverage tree to reflect all new and reorganized test cases, and clarified scenario counts and feature groupings for better maintainability and traceability.

These changes ensure more robust test coverage, better alignment between capabilities and tests, and improved clarity on API access and visibility rules across both Publisher and DevPortal.


all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/ApplicationBaseSteps.java
```diff
@@ -2493,6 +2511,34 @@ public void iSearchDevPortalAPIsWithQueryUntilContains(String query, String expe
...
+        while (true) {
+            response = Requests.get(url, headers);
+            boolean absent = response != null && response.getResponseCode() == 200
+                    && response.getData() != null && !response.getData().contains(resolvedUnexpected);
+            if (absent || System.currentTimeMillis() >= endTime) {
+                break;
+            }
+            Thread.sleep(2000);
+        }
+    }
```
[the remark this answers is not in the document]
nimsara66: @coderabbitai,

Good catch — both parts of this finding verified as valid and fixed.

1. Redundant null check: confirmed dead code — Requests.get funnels through SimpleHTTPClient's retry wrapper, which always returns a constructed response or throws IOException, so it can never return null. Removed, keeping the status-code and body checks (the body can legitimately be null).
2. Silent timeout: confirmed a real contract violation — a step named until it does not contain … within N seconds must fail on its own when the condition never holds, not rely on a following Then. Added an explicit Assert.assertTrue(absent, …) after the loop, with the query, the expected-absent value, and the last response code/body in the failure message.

Since this was a pattern issue rather than a one-off, I also swept all ~35 until…within polling steps in the glue for the same shape. Two true siblings had the identical silent-timeout + dead-null-check combination (search … until it contains and tag cloud until it contains) — fixed the same way. Everything else already asserts in-step, delegates to an asserting helper, or throws on deadline. One deliberate non-fix: the plain I search DevPortal APIs with query {string} step also retries-while-empty, but it makes no "until" promise and is used for negative searches, so an in-step assert would break legitimate absence assertions.

Validated with the focused DevPortal-search block (22/22 green; the modified steps exercised 28×/6×/2× with no spurious failures).

This is exactly the kind of finding that's most valuable here — a subtle false-green path is worse than a failing test, and catching one usually flags a family of siblings rather than a single line.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/GatewayRestArtifactsSteps.java
```diff
@@ -0,0 +1,102 @@
...
+            response = Requests.get(url, gatewayBasicAuthHeaders());
+            if (response != null && response.getResponseCode() == 200) {
+                return;
+            }
+            Thread.sleep(2000);
+        } while (System.currentTimeMillis() < endTime);
+        Assert.assertEquals(response == null ? -1 : response.getResponseCode(), 200,
+                "Gateway artifact '" + kind + "' for " + resolvedName + " did not become available within "
+                        + timeoutSeconds + "s; last: " + (response == null ? "null" : response.getData()));
+    }
```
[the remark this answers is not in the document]
nimsara66: @coderabbitai,

Verified — all three parts of this hold up against the code, and it's a sharper catch than it first looks.

1. Catch IOException in the poll loop: confirmed a real gap — SimpleHTTPClient only retries the transient 900967 "General Error" response; a network-level IOException (exactly what you'd expect from the gateway right after a deploy, which is the window this step exists for) propagated and killed the step mid-poll, defeating the retry. Now caught-and-retried per the glue's own retry-loop convention.
2. Retain the last response: done — the call sits in the try, so a throw skips the assignment and the previous real response survives into the failure message.
3. Ternary → assertNotNull + assertEquals: done. Worth noting the ternary's null arm was actually dead code before this change (Requests.get can never return null — it throws or returns a constructed response); with the catch added, all-attempts-failed becomes a genuine case, and the two-assert shape now handles it with a distinct, clear message per failure mode.

Since this is a pattern, I swept all 26 polling loops in the glue for the same shape:
- 8 identical siblings fixed (all idempotent GET polls — the DevPortal search/tag-cloud family, publisher find-by-name, endpoint-cert usage, content search) with the same mechanical treatment.
- 3 deliberately skipped: two are POST retries (API create, DCR) where blindly re-firing after a lost response isn't idempotent — a retry there risks duplicate creation and needs an already-applied check to be safe, which is a different change; and one boot-time provisioner whose fail-fast-with-raw-cause contract is the better diagnostic. Skipping those was a judgment call, happy to revisit.
- The ternary-in-asserted-value pattern had no other instances (remaining ternaries are null-safe message formatting only).

Validated with focused block runs (gateway-artifacts 2/2; search + definitions blocks covering the sibling fixes).

Findings like this one are exactly what we want surfaced — a swallowed transient on the polling path is invisible until CI load makes it intermittent

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/JwtGrantSteps.java
```diff
@@ -0,0 +1,291 @@
...
+        issuer = Utils.resolveContextPlaceholders(issuer);
+        String envelope = "<soapenv:Envelope xmlns:soapenv=\"http://schemas.xmlsoap.org/soap/envelope/\" "
+                + "xmlns:mgt=\"" + IDP_MGT_NS + "\"><soapenv:Header/><soapenv:Body>"
+                + "<mgt:getIdPByName><mgt:idPName>" + issuer + "</mgt:idPName></mgt:getIdPByName>"
+                + "</soapenv:Body></soapenv:Envelope>";
+        HttpResponse response = SimpleHTTPClient.getInstance().sendSoapRequest(
+                Utils.getIdentityProviderMgtServiceURL(getBaseUrl()), envelope, "urn:getIdPByName",
+                Identity.actingActor().getUserName(), Identity.actingActor().getPassword());
+        TestContext.set("httpResponse", response);
+    }
```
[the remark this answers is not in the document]
nimsara66: @coderabbitai,

Verified and fixed — with one deliberate deviation from the letter of the suggestion.

The issue is real: all three steps hand-wrote sendSoapRequest(...) + TestContext.set("httpResponse", ...) with no clear-before, sidestepping the funnel contract the rest of the glue relies on (clear → call → set, so a throw can never strand a stale response for a later Then to assert against).

The deviation: rather than building the Basic-Auth/SOAPAction/text-xml headers by hand and calling Requests.post in each step, the steps now call Requests.soap(...) — which already existed in the funnel class precisely for this. Same clear-before/set-after outcome, but the SOAP header construction stays centralized in SimpleHTTPClient.sendSoapRequest (which is exactly Basic + SOAPAction + text/xml under the hood) instead of being copied into the step class three times. Each SOAP envelope is preserved byte-for-byte; Requests itself needed no change.

Sibling sweep: one more step had the identical pattern — KeyManagerAdminSteps#iRevokeTheOAuthApplicationState (the updateConsumerAppState SOAP call) — fixed the same way, so 4 sites total. I also audited every other manual httpResponse publish in the glue (12 sites): they're all either correct hand-written clear/set pairs, assert-guarded poll-loop publishes (where a single-shot funnel can't apply), or custom-mTLS-client steps that can't route through the shared HTTP client — left as-is with reasons. The provisioning helpers that call SOAP and consume the response locally were left on SimpleHTTPClient directly, per the funnel's own contract.

Validated with the jwt-grant + application-revocation blocks (14/14 green; the trusted-IdP steps exercised 24×, the revoke step 2×).

Another one worth having caught — funnel bypasses are invisible until the one bad day a throw lines up with a lookalike assertion, and they tend to breed by copy-paste.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/KeyManagerAdminSteps.java
```diff
@@ -0,0 +1,227 @@
...
+        String envelope = "<soapenv:Envelope xmlns:soapenv=\"http://schemas.xmlsoap.org/soap/envelope/\" "
+                + "xmlns:xsd=\"" + ns + "\"><soapenv:Header/><soapenv:Body>"
+                + "<xsd:updateConsumerAppState><xsd:consumerKey>" + consumerKey + "</xsd:consumerKey>"
+                + "<xsd:newState>" + newState + "</xsd:newState>"
+                + "</xsd:updateConsumerAppState></soapenv:Body></soapenv:Envelope>";
+        HttpResponse response = SimpleHTTPClient.getInstance().sendSoapRequest(
+                Utils.getOAuthAdminServiceURL(getBaseUrl()), envelope, "urn:updateConsumerAppState",
+                actor.getUserName(), actor.getPassword());
+        TestContext.set("httpResponse", response);
+    }
```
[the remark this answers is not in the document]
nimsara66: @coderabbitai,

Already handled — this exact site was caught by the sibling sweep while addressing the same pattern in JwtGrantSteps (same review round): it now goes through the shared Requests.soap(...) helper with the direct sendSoapRequest call and the manual TestContext.set("httpResponse", ...) removed, and the URL/SOAPAction/envelope/credentials untouched. Validated via the application-revocation block (14/14, step exercised). Nice convergence — the review and the sweep flagging the same spot independently is good evidence the pattern hunt was complete.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/ServiceCatalogSteps.java
```diff
@@ -0,0 +1,329 @@
...
+    @When("I import a service catalog archive {string} with overwrite {string} as {string}")
+    public void iImportServiceArchive(String archiveResource, String overwrite, String idKey) throws IOException {
+        File archive = definitionFile(archiveResource);
+        Map<String, File> files = new LinkedHashMap<>();
+        files.put("file", archive);
+        TestContext.remove("httpResponse");
+        HttpResponse response = SimpleHTTPClient.getInstance().doPostMultipartWithFiles(
+                Utils.getServiceCatalogImportURL(getBaseUrl(), Boolean.parseBoolean(overwrite)),
+                serviceCatalogHeaders(), files, new HashMap<>());
+        TestContext.set("httpResponse", response);
```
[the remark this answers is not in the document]
nimsara66: @coderabbitai,

Verified and applied — both sites converted.

Fair catch on consistency: these two were actually the correct hand-written form (they did clear-before/set-after properly, unlike the SOAP sites from the earlier finding), but they were the odd ones out — the same file already used Requests.postMultipart for its two negative variants three lines away. Both imports now go through the funnel: the remove → doPostMultipartWithFiles → set triple collapses to a single Requests.postMultipart(...) call with the URL, headers, files map, and empty form-data args preserved verbatim, and the returned response still feeds the existing 200-assert / service-ID registration logic unchanged. The one remaining direct SimpleHTTPClient use in the file is the service-catalog-scoped token-mint helper (consumed locally, asserted inline, never published) — that stays, per the funnel's own contract for intermediate calls.

Sweep result: hunting the rest of this pattern turned up one more flavor — GovernanceBaseSteps had two spots that already called the funnel but kept a redundant manual TestContext.remove("httpResponse") in front of it (dead code; the funnel clears unconditionally as its first act — clearly leftovers from a pre-funnel version). Removed. With that, the glue now has zero manual httpResponse clears outside the funnels, and the only remaining manual publishes are the structurally-justified ones (assert-guarded poll-loop publishes and the mTLS client, which can't route through the shared client).

Validated with the service-catalog block (10/10 green, the import steps exercised 14× including the overwrite matrix); the governance block is running the same gate for the dead-code removal.

This one's a good example of why the "boring" consistency findings matter — every hand-rolled copy of a contract is a place for the next edit to get it subtly wrong. The funnel now owns httpResponse management uniformly, which is a strictly easier invariant to review.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/VisibilityAccessSteps.java
```diff
@@ -0,0 +1,404 @@
...
+                if (resolved.equals(list.getJSONObject(i).optString("value"))) {
+                    present = true;
+                    break;
+                }
+            }
+        }
+        Assert.assertFalse(present,
+                "Restricted-visibility tag '" + resolved + "' is present in the tag cloud but should be hidden; "
+                        + "tag cloud: " + response.getData());
+    }
```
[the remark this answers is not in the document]
nimsara66: @coderabbitai,

Verified and fixed — and it's a stronger catch than framed: on an absence assertion, the missing guards aren't just an ugly-exception risk, they're a false-pass vector. An error response's JSON has no list, the null-tolerant optJSONArray shrugs, and "restricted tag is hidden" passes vacuously against a 401/500. My earlier guard sweep had wrongly classified this step as fine — absence checks need stricter guards than presence ones, and this finding corrected that lens.

All three guards added before parsing: non-null (existing), 2xx, and non-empty body, each with code/body in the message. Safe against current usage — the preceding until it contains poll already pins a 200-with-body.

Swept all 7 absence-flavored steps for the same shape: the only two defects are both now fixed (this + the until it does not contain search poll). The rest are safe by construction — two elegantly so (the KM poll initializes its flag to the failing value; the store-user check asserts the explicit false token). One deliberate skip: the generic The response should not contain — it's contractually paired with a status assert and legitimately used on error bodies, so a 2xx guard inside it would break valid usages.

Sharpest finding of the round — it corrected the classification rule, not just the site.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/artifacts/payloads/async/asyncapiv3_invalid.yaml
```diff
@@ -0,0 +1,19 @@
...
+operations:
+  UserSignedUp:
+    action: receive
+    channel:
+      $ref: '#/channels/user~1signedup'
+components:
+  messages:
+    UserSignedUp:
+      payload:
+        type: object
```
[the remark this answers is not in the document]
nimsara66: @coderabbitai,

Verified and fixed — subtler than it looks: ~1 is JSON-Pointer escaping for /, so the old ref actually dereferenced correctly to the defined user/signedup channel. The test passed only because the parser doesn't unescape ~1 — we were pinning a parser quirk, and a future parser fix would have silently turned the fixture valid. The ref now targets a genuinely nonexistent channel, info.title is added so the fixture carries exactly one defect (no validation-order dependence), and the trailing newline is fixed. The fragility was inherited — it's a verbatim legacy copy.

Sweep: the only other message-pinned fixture (asyncapi.yaml, invalid-v2) audits clean — single, genuinely spec-required defect. The newline scan found 13 more fixture files missing EOF newlines; all fixed.

Validated: streaming-design block 18/18, invalid-v3 still failing with exactly "Operation channel reference must point to valid channel."

A negative test is only as honest as its fixture — good catch.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/features/gateway/lifecycle_stage_invocation.feature
```diff
@@ -53,3 +57,222 @@ Feature: Gateway Lifecycle-Stage Invocation
...
+    Then The response status code should be 200
+
+    # Submit an update that changes the context to a fresh unique value; redeploy so any routing change would
+    # take effect. The candidate new context is captured up-front so it can be invoked verbatim afterwards.
+    When I generate a unique value and store it as "ctxNewContext"
+    And I set the field "context" to "{{ctxNewContext}}" in the payload "ctxPayload"
+    And I update "apis" resource of id "ctxApiId" with payload "ctxPayload"
+    Then The response status code should be 200
+    And I deploy the API with id "ctxApiId"
+
```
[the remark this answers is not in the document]
nimsara66: @coderabbitai,

Valid and fixed — and this was the one place where the missing assertion was dangerous, not just untidy: the downstream check expects a 404 on the new context, which a silently failed redeploy satisfies trivially — the scenario would prove nothing about context immutability. Added the same Then …201 the feature's sibling deploys use.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/features/publisher/api_config.feature
```diff
@@ -255,3 +255,85 @@ Feature: Publisher API Runtime & Common Configuration
...
+    And I put the response payload in context as "nullApiFull2"
+    And I set the field "endpointConfig" to null in the payload "nullApiFull2"
+    And I update "apis" resource of id "nullApiId" with payload "nullApiFull2"
+    Then The response status code should be 200
+
+    Examples:
+      | actor                     |
+      | publisherUser             |
+      | publisherUser@tenant1.com |
+
```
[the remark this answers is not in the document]
nimsara66: @coderabbitai,

Applied — with the expected values corrected by live observation rather than taken from the suggestion, and the result is more interesting than either of us assumed.

The intent was right: the 200-only checks pinned null-acceptance (the NPE regression this scenario guards) but said nothing about what nulling does. However, "assert the field is effectively absent or null" turned out to be true for only one of the two fields. Probed live before asserting anything:

- securityScheme: null is ignored/re-defaulted — the retrieved API still carries ["oauth_basic_auth_api_key_mandatory","oauth2"]. Asserting "absent or null" here would have broken the scenario.
- endpointConfig: null is applied — the field is genuinely gone, production/sandbox endpoints removed.

The scenario now pins both behaviors exactly (value-match for the re-defaulted schemes; 200 + endpoints-absent for the cleared config) and documents the asymmetry in comments — a real, non-obvious product semantic that a uniform assertion would have papered over.

all-in-one-apim/modules/integration-v2/tests-common/testcontainers/src/main/java/org/wso2/am/testcontainers/DynamicApimContainer.java
```diff
@@ -211,6 +212,48 @@ public String getContainerTomlPath() {
...
+                + "org.h2.tools.RunScript -url 'jdbc:h2:./" + dbRelativePath + "' "
+                + "-user wso2carbon -password wso2carbon -script dbscripts/h2.sql";
+        Container.ExecResult r = execInContainer("bash", "-c", runScript);
+        if (r.getExitCode() != 0) {
+            throw new IOException("Secondary user-store H2 schema creation (RunScript) failed, exit "
+                    + r.getExitCode() + "\nstdout: " + r.getStdout() + "\nstderr: " + r.getStderr());
+        }
+        // execInContainer may create the file as root; the server runs as wso2carbon and must WRITE it.
+        execInContainer("bash", "-c", "chmod 0666 " + home + "/" + dbRelativePath + ".mv.db");
+        logger.info("Created secondary user-store H2 schema at {} (via product dbscripts/h2.sql)", dbRelativePath);
```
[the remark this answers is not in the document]
nimsara66: @coderabbitai,

Valid and fixed — fair cop on code written this same week: the RunScript check and the chmod were written in the same breath, and only one got the guard, even though the failure knowledge was sitting in the javadoc right above it. A failed chmod here is worse than a boot failure — H2 opens a non-writable .mv.db read-only, so store writes return 2xx while silently no-opping (isExistingUser false after a successful-looking addUser); that exact silent degradation cost real debugging time in the original fixture era, which is why the chmod exists at all. The exit code is now validated in the same pattern as the RunScript step, with exit/stdout/stderr and the silent-no-op consequence spelled out in the exception message.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/ApiProviderChangeSteps.java
```diff
@@ -0,0 +1,66 @@
...
+     * @param providerName the new provider username (may be tenant-qualified or {@code SECONDARY/}-prefixed)
+     */
+    @When("I change the provider of API {string} to {string}")
+    public void iChangeApiProvider(String apiIdKey, String providerName) throws IOException {
+        String apiId = TestContext.resolve(apiIdKey).toString();
+        String provider = Utils.resolveContextPlaceholders(providerName);
+        Map<String, String> headers = new HashMap<>();
+        headers.put(Constants.REQUEST_HEADERS.AUTHORIZATION, "Bearer " + Identity.adminToken());
+        // change-provider is a POST with the new provider as a query param and no body.
+        HttpResponse response = Requests.post(Utils.getAPIProvider(getBaseUrl(), apiId, provider), headers, "",
```
AnuGayan: 🟡 **Minor — unused local variable.** `response` is assigned but never read, and `Requests.execute()` already publishes to the shared `httpResponse` context key internally (it clears then `TestContext.set(HTTP_RESPONSE_KEY, response)`). PMD/Checkstyle will flag the dead assignment. Drop the assignment (and the now-unused `HttpResponse` import):

```java
Requests.post(Utils.getAPIProvider(getBaseUrl(), apiId, provider), headers, "",
        Constants.CONTENT_TYPES.APPLICATION_JSON);
```

The javadoc's "publishes the response as `httpResponse`" stays accurate — that happens inside `Requests`.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/ApplicationBaseSteps.java
```diff
@@ -1668,18 +1668,40 @@ public void iGetSubscription(String subscriptionId) throws Exception {
...
+     *
+     * @param appIdKey Context key containing the application ID
+     */
+    @When("I retrieve all subscriptions of application {string}")
+    public void iRetrieveAllSubscriptionsOfApplication(String appIdKey) throws IOException {
+
+        String appId = TestContext.resolve(appIdKey).toString();
+        Map<String, String> headers = new HashMap<>();
+        headers.put(Constants.REQUEST_HEADERS.AUTHORIZATION, "Bearer " + Identity.devportalToken());
+        HttpResponse response = Requests.get(
```
AnuGayan: 🟡 **Minor — unused local variable.** Same as `ApiProviderChangeSteps`: `response` is never read and `Requests.get(...)` already publishes to `httpResponse`, which the following `subscriptionShouldBeInTheListOfAllSubscriptions` step reads. Drop the `HttpResponse response =` assignment.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/PublisherBaseSteps.java
```diff
@@ -806,6 +891,33 @@ public void iPrepareDocumentOfTypeAndSource(String name, String type, String sou
...
+    public void iAddInlineDocumentContent(String content, String documentID, String apiID) throws IOException {
+
+        String docId = TestContext.resolve(documentID).toString();
+        String actualApiId = TestContext.resolve(apiID).toString();
+        Map<String, String> headers = new HashMap<>();
+        headers.put(Constants.REQUEST_HEADERS.AUTHORIZATION, "Bearer " + Identity.publisherToken());
+        Map<String, String> formFields = new HashMap<>();
+        formFields.put("inlineContent", Utils.resolveContextPlaceholders(content));
+
+        HttpResponse response = Requests.postMultipart(
```
AnuGayan: 🟡 **Minor — unused local variable.** `response` in `iAddInlineDocumentContent` is assigned but never read (`Requests.postMultipart` publishes `httpResponse` itself). Drop the assignment. (Consistent with `CommentRatingSteps`, which correctly calls `Requests.*` without capturing.)

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/utils/TenantUserProvisioner.java
```diff
@@ -221,6 +246,232 @@ public static void addRole(String tenantDomain, String roleName) throws IOExcept
...
+        for (String role : roles.split("\\s*,\\s*")) {
+            rolesXml.append("<ser:roleList>").append(role).append("</ser:roleList>");
+        }
+        // RemoteUserStoreManagerService operations live in the http://service.ws.um.carbon.wso2.org namespace —
+        // using the axis2 xsd namespace here fails with "namespace mismatch require .../service.ws.um... found
+        // .../org.apache.axis2/xsd". The addUser operation is positional: userName, credential, roleList*, claims*,
+        // profileName, requirePasswordChange.
+        String payload = "<soapenv:Envelope xmlns:soapenv=\"http://schemas.xmlsoap.org/soap/envelope/\" "
+                + "xmlns:ser=\"http://service.ws.um.carbon.wso2.org\"><soapenv:Header/><soapenv:Body>"
+                + "<ser:addUser><ser:userName>" + userName + "</ser:userName>"
```
AnuGayan: 🟡 **Minor — unescaped values in SOAP/XML envelopes.** `userName`, `password`, and role names are concatenated straight into the SOAP body here (and in `addRole`, `addSecondaryUserStore`, `getRoleListOfUser`, `isExistingUser`, `deleteRole`, `deleteUser`, plus the IdP envelopes in `JwtGrantSteps`). A value containing `&`, `<`, or `>` would corrupt the envelope. Inputs are test-controlled so real risk is low, but a small `escapeXml(...)` helper applied to interpolated values would harden these framework utilities against a future feature that uses a special char (e.g. a password with `&`).

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/BaseSteps.java
```diff
@@ -576,6 +581,20 @@ public void iSetFieldToEmptyArrayInPayload(String field, String contextKey) {
...
+     * Sets a top-level field of a JSON payload in context to JSON {@code null}, writing it back under the key. Used
+     * to prove the publisher accepts an API update that nulls an optional field (e.g. {@code securityScheme} or
+     * {@code endpointConfig}) — a past NullPointerException regression (UpdateAPINullPointerTestCase). Uses
+     * {@link JSONObject#NULL} so the field is serialized as {@code null}, not the string {@code "null"}.
+     */
+    @When("I set the field {string} to null in the payload {string}")
+    public void iSetFieldToNullInPayload(String field, String contextKey) {
+
+        org.json.JSONObject payload = new org.json.JSONObject(TestContext.resolve(contextKey).toString());
+        payload.put(field, org.json.JSONObject.NULL);
```
AnuGayan: 🔵 **Nit — import style.** `org.json.JSONObject` is used fully-qualified here (`org.json.JSONObject.NULL` and the `new org.json.JSONObject(...)` above) while the rest of the file imports `JSONObject`. Use the imported type for consistency (Checkstyle may flag).

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/BaseSteps.java
```diff
@@ -1067,33 +1093,53 @@ public void waitForAPIDeployment(String apiDetailsPayload) throws IOException, I
...
+                        && new JSONObject(revisionsResponse.getData()).getJSONArray("list").length() > 0) {
                     isApiDeployed = true;
                     break;
+                }
+                HttpResponse artifactResponse = SimpleHTTPClient.getInstance().doGet(artifactUrl, artifactHeaders);
+                if (artifactResponse != null && artifactResponse.getResponseCode() == 200) {
+                    isApiDeployed = true;
+                    break;
+                }
+            } catch (Exception e) {
```
AnuGayan: 🔵 **Nit / awareness — broad catch.** Widening to `catch (Exception e)` correctly rides out transient `JSONException` on either probe and it *is* logged with context (apiName/version/tenant), so it satisfies the no-silent-swallow rule. Just note it now also swallows programming errors (e.g. a malformed URL) until the deadline. Acceptable for a readiness loop — flagging for awareness.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/KeyManagerAdminSteps.java
```diff
@@ -0,0 +1,224 @@
...
+        org.testng.Assert.assertTrue(parts.length >= 2, "Access token is not a JWT (cannot extract claim '"
+                + claim + "'): " + token);
+        String payloadJson = new String(Base64.getUrlDecoder().decode(parts[1]), StandardCharsets.UTF_8);
+        Object value = new org.json.JSONObject(payloadJson).opt(claim);
+        org.testng.Assert.assertNotNull(value, "JWT claim '" + claim + "' not present in token payload: " + payloadJson);
+        TestContext.set(targetKey, value.toString());
+    }
+
+    /** Asserts two previously-stored context values are equal (e.g. the sub claims of two tokens). */
+    @io.cucumber.java.en.Then("The stored value {string} should equal {string}")
```
AnuGayan: 🔵 **Nit — import style.** `io.cucumber.java.en.Then`, `org.testng.Assert`, and `org.json.JSONObject` are used fully-qualified throughout this class instead of being imported. Add imports and use the short names for consistency with the other step classes.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/JwtGrantSteps.java
```diff
@@ -0,0 +1,284 @@
...
+     */
+    @When("I exchange the JWT assertion {string} for a token using consumer key {string} secret {string} scope {string}")
+    public void iExchangeJwtForToken(String jwtKey, String ckRef, String csRef, String scope) throws Exception {
+        String jwt = TestContext.resolve(jwtKey).toString();
+        String ck = TestContext.resolve(ckRef).toString();
+        String cs = TestContext.resolve(csRef).toString();
+        scope = Utils.resolveContextPlaceholders(scope);
+        String creds = Base64.getEncoder().encodeToString((ck + ":" + cs).getBytes(StandardCharsets.UTF_8));
+        StringBuilder body = new StringBuilder("grant_type=")
+                .append(java.net.URLEncoder.encode(JWT_BEARER_GRANT, StandardCharsets.UTF_8))
```
AnuGayan: 🔵 **Nit — import style.** `java.net.URLEncoder`, `java.util.Map`, and `java.util.HashMap` are used fully-qualified in this method. Import them and use the short names for consistency.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/VisibilityAccessSteps.java
```diff
@@ -0,0 +1,416 @@
...
+
+    /**
+     * Updates a published API's tags in place (GET the publisher API → replace {@code tags} → PUT) as the acting
+     * actor's publisher token. Used by ChangeAPITags to remove a tag and prove the store tag filter no longer
+     * matches it. Non-asserting — the feature asserts the PUT status.
+     */
+    @When("I set the tags of API {string} to {string}")
+    public void iSetApiTags(String apiIdKey, String tagsCsv) throws IOException {
+        JSONObject api = fetchPublisherApi(apiIdKey);
+        api.put("tags", rolesArray(tagsCsv));
```
AnuGayan: 🔵 **Nit — misleading helper name.** `rolesArray(tagsCsv)` is reused here to build the **tags** JSON array. It's functionally correct (it just trims + placeholder-resolves each CSV item into a JSON string array), but the name reads oddly for tags. Consider renaming the helper to something neutral like `csvToJsonArray`.

all-in-one-apim/modules/integration-v2/docs/devs/coverage-tree.md
```diff
@@ -2,28 +2,36 @@
 
 _Generated by `render_coverage_tree.py` from `capability-map.yml` + `.feature` tags. Do not edit by hand._ Empty branches are uncovered features (gaps to fill by eye).
 
-**299** product scenarios placed · **23** excluded (infra/framework/migration/setup) · **0** unmapped/invalid.
+**411** product scenarios placed · **23** excluded (infra/framework/migration/setup) · **0** unmapped/invalid.
```
AnuGayan: 🔵 **Nit — count mismatch with the PR description.** This generated header says **411** product scenarios placed, but the PR description states scenarios go from 299 to **399**. Please reconcile the description (the generated tree is the source of truth).

AnuGayan (commented):
Automated review of the Group-3 test port. This is a high-quality, well-documented test-only PR — defensive JSON-parse guards, correct try-with-resources, poll-on-distinguishing-state de-flaking, and genuine multi-runner awareness (unique naming, `thread-count=1` blocks for container-global mutations, poll-until-active for the async secondary-store deploy). No blockers or majors. The inline comments below are lint-level cleanups (mostly PMD/Checkstyle-style). Verdict: approve with comments.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/PublisherBaseSteps.java
nimsara66: Already done — this was one of the Requests.postMultipart sites cleaned up in the dead-variable sweep on this file, so iAddInlineDocumentContent now ends on a bare Requests.postMultipart(...) with no captured response, consistent with CommentRatingSteps.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/utils/TenantUserProvisioner.java
nimsara66: added a small Utils.escapeXml(...) (escapes & < > " ') and applied it to the interpolated data values in all the SOAP builders you listed in TenantUserProvisioner, plus the JwtGrantSteps IdP envelopes. Swept the rest of the hand-built-envelope code while here and hardened the same pattern in KeyManagerAdminSteps (updateConsumerAppState), OrganizationSteps (setUserClaimValue), and ResourceCleanup (removeOAuthApplicationData). Left base64 certs and tag-name interpolations alone (base64 has no XML metachars; a tag name can't be entity-escaped). No-op for today's test-controlled values, so the touched blocks (secondary-user-store / JwtGrant / app-revocation / org-visibility) stay green — 31/31 — while the envelopes are now robust to a future &/</> in a password, role, or URL.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/BaseSteps.java
nimsara66: Fixed, and swept the file (and the rest of the PR) for the same thing. De-qualified every org.json.* reference where the simple name was already imported — BaseSteps (6, incl. the .NULL and a stray JSONArray), ApplicationBaseSteps, ServiceCatalogSteps, PublisherBaseSteps, plus JwtGrantSteps/KeyManagerAdminSteps (added the missing import so they match the rest of the codebase). A detector now shows zero fully-qualified org.json usages left in the changed files; the one remaining sibling (MCPServerSteps) is outside this PR's file set and is tracked in the cleanup backlog.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/BaseSteps.java
nimsara66: Thanks — took it past awareness and narrowed it, since CLAUDE.md §7 makes catch-only-the-transient-set the convention for these readiness loops: the point of narrowing is exactly your observation — a programming error (a bad context key → IllegalArgumentException, or an NPE) should fail fast, not be masked as a "not deployed" timeout for the full window. Changed it to catch (IOException | JSONException e) (network warm-up + a not-yet-well-formed body — the two things that legitimately recur here) and updated the comment that claimed it "widens" the catch.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/KeyManagerAdminSteps.java
nimsara66: Done. Added the io.cucumber.java.en.Then and org.testng.Assert imports and switched this class to the short names (JSONObject was already imported from an earlier pass). Swept the other step classes for the same thing: Assert was fully-qualified throughout GovernanceBaseSteps (10) and OrganizationSteps (9) too, so those now import it as well — every step class is consistent. The one remaining sibling (MCPServerSteps) is outside this PR's file set and is tracked in the cleanup backlog.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/JwtGrantSteps.java
nimsara66: Fixed, and swept the codebase for the same pattern. Imported java.net.URLEncoder here and dropped the fully-qualified Map/HashMap (those were already imported). Same treatment for the other in-PR FQ sites: a URLEncoder/StandardCharsets helper in ApplicationBaseSteps, a stray FQ pair in Utils, and WebSocketInvocationSteps, which used FQ java.util.Map/LinkedHashMap throughout — now imports them and uses the short names like the rest of the suite. Detector confirms no fully-qualified java.net/java.util collection references remain.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/VisibilityAccessSteps.java
nimsara66: Renamed to csvToJsonArray and neutralized the parameter/locals to match — it's purely a trim + placeholder-resolve of CSV items into a JSON string array, so the neutral name fits its use for tags, visibleRoles, and accessControlRoles alike. Contained to this file; all call sites updated. Checked the other CSV-splitting spots too, but those are inline loops with purpose-specific variable names, not a reusable helper, so no similar rename needed.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/ApiProviderChangeSteps.java
nimsara66: Good catch — fixed. Requests.post() already owns the httpResponse publish (clear-before / set-after) internally, so the local was a pure dead assignment; dropped it and the now-orphaned HttpResponse import. Javadoc stays accurate since the publish still happens inside the funnel.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/ApplicationBaseSteps.java
nimsara66: Already addressed in the dead-variable sweep on this file — iRetrieveAllSubscriptionsOfApplication (the step that publishes the list subscriptionShouldBeInTheListOfAllSubscriptions asserts against) now calls Requests.get(...) bare, no HttpResponse response =. The assert step reads it back via TestContext.get("httpResponse"), so the publish contract is intact.

AnuGayan (approved):

PR #14276, opened 2026-07-27
Add a body-less SOAP invocation check to the content-aware mediation test

nisan-abeywickrama:
Integration test for the body-less SOAP fix in https://github.com/wso2/wso2-synapse/pull/2578

A SOAP envelope with no `<Body>` element, sent through a content-aware mediation flow,
used to NPE in the PassThrough sender while building the backend request — the client
got no response at all.

## What changed

One `@Test` added to the existing `OperationPolicyTestCase`:

1. Attach a content-aware mediator to the request flow.
2. POST a body-less envelope: `<soapenv:Envelope xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/"/>`
   with `content-type: text/xml` and `soapaction: "getOperation"`.
3. Assert the backend response comes back with HTTP 200.

A 30s socket timeout makes the pre-fix hang fail the test instead of blocking the suite.


nisan-abeywickrama:
Closing PR. to update the test to the new test suite

PR #14287, opened 2026-08-10
Cover Group 1 Parity

nimsara66:
This pull request introduces several new features and test runners, expands test backend capabilities, and updates documentation to clarify and support new testing scenarios. The main themes are enhancements to analytics logging, security audit, AI API, and backend test fixtures, along with improved documentation and step definitions.

**New and Enhanced Test Runners:**

* Added dedicated Cucumber test runners for correlation logging (`AnalyticsCorrelationLoggingRunner`), per-API logging (`AnalyticsPerApiLoggingRunner`), mutual SSL certificate chain validation (`GatewayMutualSslChainValidationRunner`), Gemini AI API Unlimited tier (`GatewayGeminiUnlimitedTierRunner`), devportal GraphQL complexity (`DevPortalGraphqlComplexityRunner`), and publisher security audit (`PublisherSecurityAuditRunner`). These runners isolate scenarios that require specific global configuration or overlays and port legacy test cases to the new structure. [[1]](diffhunk://#diff-3acba0354afe55f38b55545a96ead61b1b70ff727ea59f42dc8178e2f802ca17R1-R39) [[2]](diffhunk://#diff-40e1054330ae8dd619b3b567ec74fc1c7b6bed62e0fe20dc4a4f08a66447724fR1-R39) [[3]](diffhunk://#diff-bd1e98ac45e9503da55a9eee239250d6a974f2f0ee379e9c7c80119e93f78de3R1-R43) [[4]](diffhunk://#diff-2e15d8e271901774bb49bff6df9e165f7bcf20521eb95e3ee5e038b44011b6adR1-R43) [[5]](diffhunk://#diff-685f603cbb3e1984000d3436bd8b09de4da19da29fff8cd7c3ee0848efeb645aR1-R35) [[6]](diffhunk://#diff-d9245d03f933aef35aa17aefdd11be9ca2cfaa03af287f1afef6b1318fac4eb1R1-R41)

**Backend Test Fixture Improvements:**

* Added `/reflect-body-typed` route to the node backend, which echoes the request body with the same `Content-Type` as received. This enables response-flow transformation policy tests (like `xmlToJson`) to assert against XML responses.
* Added a mock Gemini AI API endpoint (`/gemini/v1beta/models/:model`) to the node backend, returning a fixed Gemini-shaped response for Unlimited tier tests.

**Capability Map and Feature Expansion:**

* Extended the capability map to include `security-audit` (API security audit report), `correlation-logging` (correlation log component configuration), and `per-api-logging` (per-API/per-resource log level) to reflect new and ported features. [[1]](diffhunk://#diff-2c58d78280c7745fb5bcccae3bd24bdfa86569609c3576eab7e0d60e9fd0332fR30) [[2]](diffhunk://#diff-2c58d78280c7745fb5bcccae3bd24bdfa86569609c3576eab7e0d60e9fd0332fR110-R111)

**Documentation Updates:**

* Clarified and expanded documentation on secondary user store provisioning, making it clear that the feature is available to any test block and describing its usage and constraints in detail.

**Step Definition Enhancements:**

* Added a new Gherkin step and implementation for invoking APIs with a credential verbatim in a specified header, supporting negative/cross-credential authentication scenarios. Refactored the retry logic to support both API keys and arbitrary credentials.


all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/features/gateway/cors.feature
```diff
@@ -329,6 +329,59 @@ Feature: Gateway CORS
...
+  #     pre-flight for POST /payee/personal answers POST-and-not-GET (the literal), while the pre-flight for
+  #     GET /payee/123 answers GET-and-not-POST (the template).
+  #   * The two resources carry different authTypes (None on the literal, Application & Application User on the
+  #     template), so the SAME path under a different verb proves the split: an unauthenticated POST
+  #     /payee/personal is 200 (literal, unsecured) while an unauthenticated GET /payee/personal is 401 (it
+  #     resolved to the secured template, not to the literal that happens to share the path).
+  # Default CORS (no corsConfiguration on the API) is required for this: with a configured CORS block the
+  # allow-methods header comes from the config, not from the resolved resource.
+  @cap:gateway @feat:cors @rule:similar-template @type:regression @dep:publisher @dep:devportal @legacy:APILoggingTest
+  Scenario Outline: A pre-flight and the real calls resolve to the right sibling when a literal resource sits next to a template as <actor>
```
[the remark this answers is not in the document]
nimsara66: @coderabbitai,

Can you check again.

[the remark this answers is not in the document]
nimsara66: @coderabbitai,

If this is addresses can we resolve?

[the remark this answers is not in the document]
nimsara66: @coderabbitai resolve

all-in-one-apim/pom.xml
```diff
@@ -1311,7 +1311,7 @@
         <carbon.apimgt.ui.version>9.3.194</carbon.apimgt.ui.version>
 
         <!-- APIM Component Version -->
-        <carbon.apimgt.version>9.33.162</carbon.apimgt.version>
+        <carbon.apimgt.version>9.33.168</carbon.apimgt.version>
```
AnuGayan, line 1314: _📦 Build & Dependencies_ | _🟠 Major_ | _🔍 Needs discussion_

**Product dependency bump riding inside a test-coverage PR.**

`carbon.apimgt.version` moves `9.33.162` → `9.33.168` here and in the three sibling poms (`api-control-plane/pom.xml:1288`, `gateway/pom.xml:1288`, `traffic-manager/pom.xml:1290`), which changes what all four distributions build and ship. Nothing in the PR description mentions it, so it cannot be reviewed alongside a 168-file test change and it will not be findable by a later bisect.

Either split it into its own PR, or state in the description which new scenario needs which fix between 163 and 168 — e.g. the `businessPlan` query parameter or the `regenerate-secret` path that `IntegrationV2-SingleClientSecret` depends on. If the suite is green on `.162`, the bump should go separately.

all-in-one-apim/modules/integration-v2/docs/devs/coverage-tree.md
```diff
@@ -2,60 +2,67 @@
 
 _Generated by `render_coverage_tree.py` from `capability-map.yml` + `.feature` tags. Do not edit by hand._ Empty branches are uncovered features (gaps to fill by eye).
 
-**594** product scenarios placed · **31** excluded (infra/framework/migration/setup) · **0** unmapped/invalid.
+**708** product scenarios placed · **26** excluded (infra/framework/migration/setup) · **0** unmapped/invalid.
```
AnuGayan, line 5: _📄 Docs & Generated Artifacts_ | _🟡 Minor_ | _⚡ Quick win_

**The regenerated coverage tree is stale — it predates the last feature edits.**

Running the repo's own generator against this branch produces a different file:

```
$ python3 all-in-one-apim/modules/integration-v2/docs/devs/render_coverage_tree.py
capabilities: 7 | scenarios: 735 | placed: 709 | excluded: 26 | invalid: 0
```

The committed header says **708**. The delta is 197 lines: ~196 stale line numbers (e.g. `publisher/api_lifecycle.feature:167` → `:170`, and most of `publisher/api_config.feature`) plus one scenario missing from the tree entirely — `[unregistered-scope] A scope not registered on the application is still granted in the issued token` (`key-manager/token_issuance.feature:420`), which this PR adds.

Tag lint itself is clean (`invalid: 0`), so this is purely a re-render: run the script and commit the result. Worth noting the script exits non-zero on invalid tags and is explicitly designed as a CI gate — wiring it in would stop this drifting again.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/MutualSslSteps.java
```diff
@@ -19,6 +19,7 @@
 
 import io.cucumber.java.en.When;
 import org.wso2.am.integration.cucumbertests.utils.Identity;
+import org.wso2.am.integration.cucumbertests.utils.JwtTestUtils;
```
AnuGayan: _🧹 Maintainability_ | _🟡 Minor_ | _⚡ Quick win_

**Unused import.**

`JwtTestUtils` is imported but never referenced anywhere in the file — the new `bearer(...)`, `uploadClientCertificate(...)` and `invokeMtlsUntilStatus(...)` members do not use it. Checkstyle/PMD (`UnusedImports`) will flag it.

<details>
<summary>🛡️ Proposed fix</summary>

```diff
 import org.wso2.am.integration.cucumbertests.utils.Identity;
-import org.wso2.am.integration.cucumbertests.utils.JwtTestUtils;
 import org.wso2.am.integration.cucumbertests.utils.Requests;
```
</details>

AnuGayan (commented):
## Review — integration-v2 Group 1 parity

Reviewed the full diff (168 files, +14,775/−1,204) against head `1ee503f`, with the step-definition/feature wiring and the coverage-tree generator checked mechanically rather than by eye. These comments are additive to the CodeRabbit thread above — no overlap with the items already addressed in `4db3321`.

**Overall: approve with comments.** The test code itself has no blockers. The engineering quality is high and unusual in a good way: the co-hosting and `thread-count` decisions are argued in place with measured evidence, and several changes fix scenarios that were previously green *for the wrong reason*. Two items are outside that: a product dependency bump that does not belong to this PR's stated scope, and a generated doc that is out of sync with its generator.

### Needs an answer before merge
1. **`carbon.apimgt.version` 9.33.162 → 9.33.168** across all four distribution poms, unmentioned in the description — split it out or say which scenario needs which fix (comment on `all-in-one-apim/pom.xml`).
2. **`docs/devs/coverage-tree.md` is stale**: re-rendering yields 709 placed scenarios vs the committed 708, missing this PR's own `[unregistered-scope]` scenario plus ~196 drifted line numbers. One command to fix.

### Not applicable
No DDL, migrations, ORM models, or product endpoints change here, so the database-query, schema/data-type and REST-contract dimensions have nothing to flag. The one contract-adjacent change is a client-side *fix* — see below.

### Verified mechanically
- **Step wiring is clean.** No duplicate step-definition texts across 785 annotations; every changed step text has all its feature call sites migrated (no orphaned `allowing role`, no old `expecting error code {int} within` phrasing); the 28 mutual-SSL fixture references all resolve to files that exist after the `.jks` → `.p12` renames.
- **Tag/capability-map consistency.** `render_coverage_tree.py` reports `invalid: 0`, and the three new `capability-map.yml` feature entries match live `@feat:correlation-logging` / `@feat:per-api-logging` / `@feat:security-audit` tags.
- **One dead step**, flagged inline (`UserAccountSteps:106`).

### What's good
- **Fixes to tests that were passing for the wrong reason.** `getChangeSubscriptionBusinessPlanURL` sent `throttlingPolicy` where publisher-api.yaml declares `businessPlan`, so every 400 assertion against that endpoint was satisfied by a missing-parameter error. Same class: the graphql-ws query-param arc was sending a literal `Bearer null`, so `AUTH_IN.QUERY` was never exercised; the `X-WSO2-CLIENT-CERTIFICATE` negative was satisfied by a base64url decode failure rather than by the rule under test.
- **`SimpleHTTPClient.keystoreTypeOf`** — magic-byte detection instead of hardcoded `"JKS"`, plus renaming the fixtures to `.p12` to match what they always were (rename-only, content untouched). Removes a silent dependency on the JDK `keystore.type.compat` shim.
- **`registerIfCreated`** applied uniformly to every `I attempt to …` create step, with the no-op-on-rejection contract preserved. Closes a real leak class in shared containers.
- **`userExists` tri-state** — refusing to read a SOAP fault as "already absent" is exactly the failure mode that makes a leak look like a clean teardown. Same instinct in the isolated IS sweep.
- **Assertions built to be non-vacuous**: `definitionShouldNotExposePublisherOnlyExtensions` deriving its expectation set from the publisher copy and failing as `UNEXERCISED` otherwise; the `operations.length() > 0` guards ahead of every "every operation …" claim; `assertSameField` requiring key presence so `null == null` cannot pass.
- **The verify-first commentary.** Nearly every non-obvious choice cites the product call site or is marked as measured — why the MCP server-level policy must be JSON `null`, why the multi-group extractor returns the bare claim, why the rotation endpoint only works under the single-secret overlay. That, plus the block rationale in `testng-v2.xml`, is what makes a change this size reviewable at all.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/UserAccountSteps.java
```diff
@@ -0,0 +1,143 @@
...
+     * partial version of it.
+     *
+     * @param namePrefix  prefix for the generated local part (e.g. {@code apisecEmailUser})
+     * @param emailDomain the domain that follows the {@code @} INSIDE the username (e.g. {@code wso2.com}) — this is
+     *                    part of the name, and is unrelated to the tenant the user lives in
+     * @param password    the user's initial password
+     * @param roles       comma-separated Internal roles (e.g. {@code Internal/subscriber})
+     * @param usernameKey context key for the store name; the credential name goes to {@code <usernameKey>LoginName}
+     */
+    @When("I provision a user with name prefix {string} and email domain {string} password {string} and roles {string} storing the username as {string}")
```
AnuGayan, line 138: _🧹 Maintainability_ | _🟡 Minor_ | _🔍 Needs decision_

**This step definition has no call site — it is dead glue.**

No feature file uses it:

```
$ grep -rn "email domain" src/test/resources/features
(no matches)
```

That is ~35 lines of implementation plus a detailed Javadoc explaining the doubly-qualified `user@emailDomain@tenantDomain` credential rule and the `BasicAuthAuthenticator` last-`@` split — a genuinely useful finding that nothing currently exercises, so it cannot regress if the rule changes. The sibling `iProvisionAScenarioUser` **is** used, so this is specific to the email-form variant.

Either land the scenario that consumes it in this PR (the Javadoc points at `APISecurityTestCase`'s `apisecUser2@wso2.com` / `@abc.com` cases), or drop the step and keep the note in the parity backlog until that scenario arrives.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/OasDefinitionSteps.java
```diff
@@ -0,0 +1,702 @@
...
+    /**
+     * Asserts two stored definitions are IDENTICAL, not merely operation-compatible — a semantic (canonicalised)
+     * comparison via {@code JSONObject.similar}, so a re-serialisation that reorders keys is not a difference.
+     *
+     * <p>The operation-level check alone cannot see a rejected update that still corrupted the document on the way
+     * out: descriptions, {@code x-} extensions, endpoint config and {@code servers} all live outside {@code paths}.
+     * On failure the differing top-level sections are named, because a raw dump of two OpenAPI documents is
+     * unreadable.</p>
+     */
+    @Then("The definitions stored as {string} and {string} should be identical")
```
AnuGayan, line 520: _🧹 Maintainability_ | _🟡 Minor_ | _⚡ Quick win_

**Javadoc is attached to the wrong method.**

The `@Then(... should be identical)` annotation is followed by a blank line and then the Javadoc block that documents `definitionsShouldDeclareTheSameOperations` ("the SAME resource surface … ports legacy's `validateUpdatedDefinition`"), and only then by `definitionsShouldBeIdentical`. The annotation still binds correctly — comments are ignored — but the net effect is that `definitionsShouldBeIdentical` is documented as something it does not do, and `definitionsShouldDeclareTheSameOperations` (line 553) has no doc at all. Checkstyle's Javadoc placement rules will also object to a comment between an annotation and its member.

Move the block down to sit directly above `definitionsShouldDeclareTheSameOperations`, and give `definitionsShouldBeIdentical` its own one-paragraph doc (the "semantic `JSONObject.similar` comparison, names the differing top-level sections" paragraph that currently opens the misplaced block).

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/OasDefinitionSteps.java
```diff
@@ -0,0 +1,702 @@
...
+            } else if (a instanceof JSONArray && b instanceof JSONArray) {
+                same = ((JSONArray) a).similar(b);
+            } else {
+                same = a.equals(b);
+            }
+            if (!same) {
+                differing.add(section);
+            }
+        }
+        Assert.fail("The stored definition changed across the rejected update: section(s) " + differing
```
AnuGayan: _🧹 Maintainability_ | _🔵 Nit_ | _⚡ Quick win_

**Generic step, caller-specific failure text — and the before/after values print empty.**

Two small things in the same message:

1. "changed across the rejected update" is one caller's narrative; the step itself just compares two stored definitions, so a failure from any other call site reads as a claim about an update that never happened.
2. `optString(section)` returns `""` for a section whose value is a `JSONObject`/`JSONArray` — which is every section this branch can report (`paths`, `info`, `servers`, `x-wso2-*`). So `before=`/`after=` are always blank for exactly the sections that differ.

<details>
<summary>🛡️ Proposed fix</summary>

```diff
-        Assert.fail("The stored definition changed across the rejected update: section(s) " + differing
-                + " differ between '" + firstKey + "' and '" + secondKey + "'."
-                + " before=" + first.optString(differing.isEmpty() ? "info" : differing.get(0))
-                + " after=" + second.optString(differing.isEmpty() ? "info" : differing.get(0)));
+        String section = differing.isEmpty() ? "info" : differing.get(0);
+        Assert.fail("The definitions stored as '" + firstKey + "' and '" + secondKey + "' are not identical:"
+                + " section(s) " + differing + " differ."
+                + " '" + section + "' before=" + String.valueOf(first.opt(section))
+                + " after=" + String.valueOf(second.opt(section)));
```
</details>

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/ApplicationBaseSteps.java
```diff
@@ -1785,7 +2167,7 @@ private JSONObject reflectedApplicationAttributes() {
     @Then("The reflected backend JWT applicationAttributes claim should contain {string} with an empty value")
     public void theReflectedBackendJwtAttributeIsEmpty(String attributeName) {
         JSONObject attrs = reflectedApplicationAttributes();
-        Assert.assertTrue(attrs.has(attributeName) && attrs.optString(attributeName).isEmpty(),
+        Assert.assertTrue(attrs.has(attributeName) && attrs.optString(attributeName).isBlank(),
```
AnuGayan: _🎯 Functional Correctness_ | _🟡 Minor_ | _⚡ Quick win_

**The `isEmpty()` → `isBlank()` sweep weakens this assertion.**

The sweep is right almost everywhere in this PR (guards on response bodies, where whitespace-only is as useless as empty), but here emptiness *is* the contract: the step is named `theReflectedBackendJwtAttributeIsEmpty` and its Gherkin says "with an empty value". After the change a claim value of `" "` satisfies it, so a backend JWT carrying a whitespace application attribute now passes as "empty".

<details>
<summary>🛡️ Proposed fix</summary>

```diff
-        Assert.assertTrue(attrs.has(attributeName) && attrs.optString(attributeName).isBlank(),
+        Assert.assertTrue(attrs.has(attributeName) && attrs.optString(attributeName).isEmpty(),
```
</details>

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/PublisherBaseSteps.java
```diff
@@ -2261,9 +2633,9 @@ public void theExportedApiArchiveShouldStripSecrets(String archivePathKey) throw
         String sandboxPassword = endpointSecurity.getJSONObject("sandbox").optString("password", "");
         // Assert the password was stripped WITHOUT echoing the value — a non-empty value here is a real backend
         // credential and must not be printed into CI output (the very leak this test guards against).
-        Assert.assertTrue(productionPassword.isEmpty(),
+        Assert.assertTrue(productionPassword.isBlank(),
```
AnuGayan: _🔐 Security & Secrets_ | _🟡 Minor_ | _⚡ Quick win_

**Same `isBlank()` widening, in the assertion that guards a credential leak.**

This step exists to prove the exporter strips endpoint passwords, and the comment two lines up is explicit that a non-empty value here is a real backend credential. With `isBlank()` a password of `" "` (or `"\n"`) now reads as "stripped", which is the one direction this assertion must not tolerate. Applies to both lines (production at 2636, sandbox at 2638).

<details>
<summary>🛡️ Proposed fix</summary>

```diff
-        Assert.assertTrue(productionPassword.isBlank(),
+        Assert.assertTrue(productionPassword.isEmpty(),
                 "Production endpoint password was exported in plain text (expected empty)");
-        Assert.assertTrue(sandboxPassword.isBlank(),
+        Assert.assertTrue(sandboxPassword.isEmpty(),
                 "Sandbox endpoint password was exported in plain text (expected empty)");
```
</details>

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/ApplicationBaseSteps.java
```diff
@@ -2060,16 +2493,16 @@ public void iCreateIsUserWithRole(String usernameBase, String password, String i
         ISResourceCleanup.registerUser(userId);
 
         // Empty key => create the user with NO role (the "without the mapped role" actor).
-        if (isRoleKey == null || isRoleKey.trim().isEmpty()) {
+        if (isRoleKey == null || isRoleKey.trim().isBlank()) {
```
AnuGayan: _🧹 Maintainability_ | _🔵 Nit_ | _⚡ Quick win_

**`trim()` is redundant before `isBlank()`.**

`isBlank()` already returns true for whitespace-only strings, so `x.trim().isBlank()` is exactly `x.isBlank()`. Introduced by the sweep at four sites:

- here (`isRoleKey.trim().isBlank()`)
- `JwtGrantSteps.java:99` (`extraClaimsJson.trim().isBlank()`)
- `JwtGrantSteps.java:175` (`scope.trim().isBlank()`)
- `utils/Utils.java:723` (`lifecycleChecklist.trim().isBlank()`)

Dropping the `trim()` keeps the intent obvious (the old `trim().isEmpty()` pairing was load-bearing; the new one is not).

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/utils/ResourceCleanup.java
```diff
@@ -195,7 +233,13 @@ public static void deleteRegisteredResources() {
         // External-system resources first (independent of the APIM lists and of APIM actor/token resolution):
         // IS-side resources are swept as the IS integration actor via IS's own management APIs — an APIM actor
         // token cannot address them. See ISResourceCleanup / CLAUDE.md §14.
-        ISResourceCleanup.sweep();
+        // Isolated: a Throwable escaping the IS sweep would skip every APIM delete below. Logged, not swallowed (§5).
+        try {
+            ISResourceCleanup.sweep();
+        } catch (Throwable t) {
```
AnuGayan: _🩺 Stability & Availability_ | _🟡 Minor_ | _⚡ Quick win_

**`catch (Throwable)` — PMD will flag it, and it downgrades `Error` to a WARN.**

Isolating the IS sweep so it cannot skip the APIM deletes is the right call, and the comment justifies it well. The catch is just too wide: `AvoidCatchingThrowable` is on by default, and an `OutOfMemoryError` / `StackOverflowError` raised during teardown now becomes a log line and a "successful" cleanup.

`Exception | AssertionError` covers everything the sweep realistically throws — the same pair `sweepUsers` already catches a few lines below, so the two stay consistent.

<details>
<summary>🛡️ Proposed fix</summary>

```diff
         try {
             ISResourceCleanup.sweep();
-        } catch (Throwable t) {
+        } catch (Exception | AssertionError t) {
             logger.warn("External IS sweep failed; IS-side resources may have leaked. Continuing with the APIM sweep "
                     + "so APIM resources are still removed.", t);
         }
```
</details>

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/LoggingSteps.java
```diff
@@ -0,0 +1,425 @@
...
+        }
+        String content = readLogFile(fileName);
+        int from = Integer.parseInt(mark.toString());
+        // The file rolls over at 10MB, which resets its length; treat a shrink as "read it all".
+        return from <= content.length() ? content.substring(from) : content;
+    }
+
+    private String readLogFile(String fileName) {
+        return container().readContainerFile(Constants.APIM_CONTAINER_USER_HOME + "/"
+                + System.getProperty("apim.server.name") + "/repository/logs/" + fileName);
```
AnuGayan: _🧹 Maintainability_ | _🟡 Minor_ | _🔍 Small refactor_

**Re-implements a container path convention the container class already owns, and degrades silently when the system property is unset.**

`DynamicApimContainer.getContainerLog4j2Path()` (line 368 of that class) assembles this exact `APIM_CONTAINER_USER_HOME + "/" + System.getProperty("apim.server.name") + "/repository/…"` shape, so the in-container layout is now encoded in two modules — and `RemoteLoggingSteps` reaches it through the accessor while this class rebuilds it by hand.

Second, `apim.server.name` is unvalidated here. `CoverageSupport:113-119` deliberately fails fast with a message naming `-Dapim.server.name=<name>` when it is missing; here an unset property yields `/…/null/repository/logs/api.log` and surfaces later as a generic `Failed to read container file`.

Suggest adding a sibling accessor next to `getContainerLog4j2Path()` and calling it:

<details>
<summary>🛡️ Proposed fix</summary>

```diff
+    /** In-container path of one of the running server's log files (for the logging assertions). */
+    public String getContainerLogFilePath(String fileName) {
+        return Constants.APIM_CONTAINER_USER_HOME + "/" + requireServerName() + "/repository/logs/" + fileName;
+    }
```
```diff
     private String readLogFile(String fileName) {
-        return container().readContainerFile(Constants.APIM_CONTAINER_USER_HOME + "/"
-                + System.getProperty("apim.server.name") + "/repository/logs/" + fileName);
+        return container().readContainerFile(container().getContainerLogFilePath(fileName));
     }
```
</details>

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/LoggingSteps.java
```diff
@@ -0,0 +1,425 @@
...
+     * the traffic that would produce the lines is driven ONCE, before the check — by the time a poll noticed,
+     * the lines would already have been written.
+     */
+    @When("I wait {int} seconds for the correlation configuration to reach the gateway")
+    public void iWaitForCorrelationConfigToPropagate(int seconds) throws InterruptedException {
+        Thread.sleep(seconds * 1000L);
+    }
+
+    /** Waits for a per-API logging event to reach the gateway after the devops PUT is acknowledged. */
+    @When("I wait {int} seconds for the per-API logging configuration to reach the gateway")
```
AnuGayan: _🧹 Maintainability_ | _🔵 Nit_ | _⚡ Quick win_

**Two step definitions with byte-identical bodies.**

`iWaitForCorrelationConfigToPropagate` (line 177) and `iWaitForPerApiLoggingConfigToPropagate` (here) are both `Thread.sleep(seconds * 1000L)` and differ only in Gherkin phrasing. Across the two features that is 8 call sites × 15 s ≈ 2 minutes of unconditional sleep in a `thread-count=1` block.

The sleeps themselves I would keep — the Javadoc's reasoning is right, the provoking traffic is driven once *after* the wait, so polling the assertion cannot replace it. Just collapse the two into one parameterised step so the phrasing stays readable without duplicating the body:

```gherkin
When I wait 15 seconds for the "correlation" configuration to reach the gateway
When I wait 15 seconds for the "per-API logging" configuration to reach the gateway
```

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/MCPInvocationSteps.java
```diff
@@ -408,6 +414,208 @@ public void invokeMcpToolExpectStatus(String toolName, String argsJson, String c
...
+        HttpClient client = newClient();
+        AtomicReference<String> lastError = new AtomicReference<>();
+        AtomicInteger lastStatus = new AtomicInteger(-1);
+        long startedMillis = System.currentTimeMillis();
+        JSONObject actual = Utils.retryUntil(timeoutSeconds * 1000L, () -> {
+            try {
+                HttpResponse<String> initResp = post(client, mcpUrl, token, null, INIT);
+                lastStatus.set(initResp.statusCode());
+                String sessionId = initResp.headers().firstValue("mcp-session-id").orElse(null);
+                if (initResp.statusCode() != 200 || !sseOrJson(initResp.body()).contains("serverInfo")) {
```
AnuGayan, line 514: _🩺 Stability & Availability_ | _🟡 Minor_ | _⚡ Quick win_

**An NPE here escapes the retry envelope the rest of the method carefully builds.**

`sseOrJson(...)` can return null, and the surrounding `catch (IOException | JSONException transientDuringWarmup)` will not catch the resulting `NullPointerException` — so a null/empty initialize body during warm-up aborts the step with an opaque NPE instead of being polled through. This is exactly the failure mode the `tools/list` half of this PR fixes (`if (listBody == null || listBody.isBlank())`) and that `mcpInitializeShouldBeExactly` guards (`body == null || body.isBlank()`).

Same line appears in the other new method at 662 (`mcpToolShouldNotBeThrottled` handshake); it is also pre-existing at 100/160/222/292, so a small shared helper would fix all six at once.

<details>
<summary>🛡️ Proposed fix</summary>

```diff
-                if (initResp.statusCode() != 200 || !sseOrJson(initResp.body()).contains("serverInfo")) {
+                String initBody = sseOrJson(initResp.body());
+                if (initResp.statusCode() != 200 || initBody == null || !initBody.contains("serverInfo")) {
                     lastError.set("init status=" + initResp.statusCode() + " body=" + initResp.body());
                     return null;
                 }
```
</details>

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/ApiProductSteps.java
```diff
@@ -0,0 +1,496 @@
...
+        Requests.postMultipart(Utils.getValidateOpenAPIURL(Utils.getBaseUrl()), headers, files, new HashMap<>());
+    }
+
+    /**
+     * Asserts a JSON ARRAY field of the last response holds exactly {@code expectedCount} entries. The count-only
+     * counterpart of the shared exact-list assertion, for a collection whose SIZE is the contract while its
+     * element shape is not a flat list of scalars — e.g. a lifecycle state's {@code availableTransitions}
+     * (objects), of which a PUBLISHED API product must offer exactly four.
+     */
+    @Then("The response array field {string} should have exactly {int} entries")
```
AnuGayan: _🧹 Maintainability_ | _🔵 Nit_ | _🔍 Small refactor_

**Generic response assertion living in product-specific glue.**

`The response array field {string} should have exactly {int} entries` reads only `httpResponse` and `Utils.extractValueFromPayload` — there is nothing API-product-specific about it, and its own Javadoc calls it "the count-only counterpart of the shared exact-list assertion". That sibling (`The response field {string} should be exactly the list {string}`) was added to `BaseSteps` in this same PR, and `BaseSteps.theResponseArrayFieldShouldBeExactly` even points back here for the count.

Moving it to `BaseSteps` keeps the pair together and stops the next author from looking for it in the product class. The class comment's rule ("kept in its own class so the shared step classes stay untouched") is about product HTTP flows, which this is not.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/BaseSteps.java
```diff
@@ -780,6 +823,127 @@ private void assertResponseFieldValue(String fieldName, String expectedValue, bo
...
+     * {@code client_secret} for Okta, {@code password} AND {@code client_secret} for PingFederate), all of
+     * which must come back as exactly the {@code *****} sentinel.
+     *
+     * @param fieldNamesCsv comma-separated field names / JSONPaths to read from the response body
+     * @param expectedValue the exact expected value every one of them must equal (string form)
+     */
+    @Then("Each of the response fields {string} should be {string}")
+    public void eachOfTheResponseFieldsShouldBe(String fieldNamesCsv, String expectedValue) throws IOException {
+
+        for (String fieldName : fieldNamesCsv.split(",")) {
```
AnuGayan, line 874: _🎯 Functional Correctness_ | _🔵 Nit_ | _⚡ Quick win_

**Splitting on `,` silently rules out comma-bearing JSONPaths.**

Fine for the current call sites (the masked connector-secret field names), but the singular sibling `theValueOfResponseFieldShouldBe` accepts a full JSONPath, and predicates routinely contain commas — `list[?(@.a=='x,y')]`, `paths['/a,b']`. Such a path would be split mid-expression and fail as "field not found", which is a confusing way to learn about the limitation.

A one-line Javadoc caveat is enough ("field names only — a JSONPath containing a comma cannot be expressed here; use the singular step per field"), no behaviour change needed.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/artifacts/configFiles/is7txSecurityAudit/deployment.toml
```diff
@@ -0,0 +1,52 @@
...
+# base_url points at the `am-auditApi-sample` Express mock baked into the node-app-server image
+# (nodeapps/am-auditApi-sample, pm2 app on port 3002, network alias `nodebackend`) — hence initBackend is
+# load-bearing for this runner. The mock speaks plain HTTP, which the product accepts: APIUtil
+# .getHttpClient() short-circuits to a plain PlainConnectionSocketFactory for non-https schemes.
+#
+# api_token is echoed as the X-API-KEY header on all three outbound calls and collection_id as the `cid`
+# multipart field on the create call. The mock checks neither, but the product requires both to be
+# non-empty or it treats the feature as unconfigured and returns null.
+[security_audit]
+api_token = "b57973cf-b74c-4ade-921d-ece83251eceb"
```
AnuGayan: _🔐 Security & Secrets_ | _🟡 Minor_ | _⚡ Quick win_

**Mock credential shaped exactly like a real one — Gitleaks/secret scanners will flag it.**

A UUID-valued `api_token` (and `collection_id`) next to a `base_url` is the canonical secret-scanner signature, and both are now committed. The comment three lines above already establishes that neither value is validated: *"The mock checks neither, but the product requires both to be non-empty."*

So make them obviously synthetic, which keeps the product happy and removes the scanner noise (and stops a future reader from wondering whether a live audit-service token leaked into the tree):

<details>
<summary>🛡️ Proposed fix</summary>

```diff
-api_token = "b57973cf-b74c-4ade-921d-ece83251eceb"
-collection_id = "f73b8171-4f71-499b-891a-d34aa71f2d45"
+api_token = "mock-audit-api-token"
+collection_id = "mock-audit-collection-id"
```
</details>

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/testng-v2.xml
```diff
@@ -550,32 +578,120 @@
...
+    <!-- thread-count=1 is REQUIRED here, not a tuning choice.
+         AnalyticsCorrelationLoggingRunner mutates the server-wide correlation logging configuration (the devops
+         /correlation config): ONE object per server, listing the five declared components (http, jdbc, ldap,
+         synapse, method-calls) and their enabled flags. Its scenarios enable a specific component set, invoke,
+         then assert the EXACT enabled set — so a class running concurrently that toggles the same config flips a
+         flag mid-scenario and the assertion fails against state it did not set.
+         Unlike a policy listing, this cannot be scoped or filtered per scenario: there is exactly one config
+         object and no tenant/resource dimension to isolate on. Serialising the block's classes is therefore the
+         only available isolation. Adding a class here that touches correlation config has the same problem. -->
+    <test name="IntegrationV2-Logging" parallel="classes" thread-count="1">
```
AnuGayan, line 650: _📄 Docs & Generated Artifacts_ | _🔵 Nit_ | _⚡ Quick win_

**The same thread-count=1 argument is made twice, and there are stray blank runs around this block.**

The new `IntegrationV2-Logging` entry is preceded by two consecutive comment blocks (~40 lines) that both argue the same conclusion: the first covers "the restart plus container-global component switches", the second ("thread-count=1 is REQUIRED here, not a tuning choice") re-derives the global-switch half in more detail. Merging them into one block — restart reason, then global-switch reason, then the "adding a class that touches correlation config has the same problem" warning — keeps the reasoning without the reader having to diff two versions of it.

Also, three consecutive blank lines were introduced before this block and after the `SolaceAndMtlsChain` block; worth collapsing to one while you are in here.

Everything else in this re-organisation reads well — the co-hosting notes stating *why concurrency is safe* separately from *why co-hosting is safe* (the `(a)`/`(b)` split) is a real improvement over what was there.

all-in-one-apim/pom.xml
nimsara66: None of the tests are dependent on the version bump. All the CI checks passing is a green flag to proceed. Hence omitting the suggested changes

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/UserAccountSteps.java
nimsara66: Fixed — the step is no longer dead: gateway/basic_auth_email_username.feature now consumes it at 6 call sites across both email domains (wso2.com and abc.com), so the doubly-qualified user@emailDomain@tenantDomain rule and the BasicAuthAuthenticator last-@ split are now exercised and will regress loudly if the rule changes.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/OasDefinitionSteps.java
nimsara66: Fixed — the block now sits directly above definitionsShouldDeclareTheSameOperations (577) and definitionsShouldBeIdentical (520) has its own doc covering the semantic JSONObject.similar comparison and the named differing sections; no comment now separates an annotation from its member, so the Checkstyle placement objection is gone too.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/MCPInvocationSteps.java
nimsara66: Not valid — sseOrJson can't return null (if (body == null) return ""; is its first statement, pre-existing in HEAD), so .contains(...) can't NPE; a null/empty init body gives "".contains("serverInfo") == false, which takes the lastError.set(...); return null; branch and is polled through the retry envelope, so the proposed initBody == null check would be permanently dead code.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/BaseSteps.java
nimsara66: Done — the Javadoc now carries a dedicated "FIELD NAMES ONLY" paragraph naming the limitation, both of your example paths (list[?(@.a=='x,y')], paths['/a,b']), the "split mid-expression → field not found" symptom, and the singular-step workaround; @param updated to "comma-free JSONPaths". No behaviour change, as you suggested.

AnuGayan (approved):

PR #14288, opened 2026-08-10
Cover Group 3 Parity

nimsara66:
This pull request introduces several improvements and new features to the integration test backend infrastructure, focusing on more accurate test coverage reporting and expanding the available backend services for testing. The most significant changes are the enhancement of the coverage tree tool to verify actual test execution, and the addition of new mock backend services (including a custom status-message backend). These updates help ensure that only scenarios actually executed by the test suite are counted as covered, and provide new test endpoints for more comprehensive integration testing.

**Test Coverage Reporting Improvements:**

* The `render_coverage_tree.py` tool now verifies that each scenario is actually executed by checking for a registered runner in `testng-v2.xml`, and excludes scenarios that are not executed from the coverage count. Unexecuted scenarios are clearly reported in the output, preventing false positives in coverage metrics. [[1]](diffhunk://#diff-6a3ebcb535f1f08f93abc37865e2b535d1ad496539f1dd1d6ddb9c1b62e5fe33R11-R51) [[2]](diffhunk://#diff-6a3ebcb535f1f08f93abc37865e2b535d1ad496539f1dd1d6ddb9c1b62e5fe33R101-R161) [[3]](diffhunk://#diff-6a3ebcb535f1f08f93abc37865e2b535d1ad496539f1dd1d6ddb9c1b62e5fe33L121-R213) [[4]](diffhunk://#diff-6a3ebcb535f1f08f93abc37865e2b535d1ad496539f1dd1d6ddb9c1b62e5fe33R247-R271) [[5]](diffhunk://#diff-6a3ebcb535f1f08f93abc37865e2b535d1ad496539f1dd1d6ddb9c1b62e5fe33R288-R304) [[6]](diffhunk://#diff-6a3ebcb535f1f08f93abc37865e2b535d1ad496539f1dd1d6ddb9c1b62e5fe33R313-R323) [[7]](diffhunk://#diff-6a3ebcb535f1f08f93abc37865e2b535d1ad496539f1dd1d6ddb9c1b62e5fe33R338-R367)

**New/Updated Mock Backends:**

* Added a new `custom-status-backend` Node.js service that returns a non-standard HTTP status reason phrase, used to test gateway behavior with custom backend responses. This service is built using raw sockets and is included in the Docker build and PM2 process manager configuration. [[1]](diffhunk://#diff-b3fc876a05b3c59e2e4b811b4c513d34010188e202b61a62cb695639e434eabaR1-R9) [[2]](diffhunk://#diff-0ec4402a52b138df0b0adda682db1e8051999a4f12d8d46e23d037f70a2b036cR1-R65) [[3]](diffhunk://#diff-fe6aa8dd4ae1c62d32ab1bd7fae58a5cb93f5bc30dc8572b7a1644c25e0aae79R31-R41) [[4]](diffhunk://#diff-fe6aa8dd4ae1c62d32ab1bd7fae58a5cb93f5bc30dc8572b7a1644c25e0aae79R69-R77) [[5]](diffhunk://#diff-1dffc1c39ecb42db30da4b9d7104255fa3522844c40010017608480833902686R170-R185)
* Updated the test container configuration to expose additional ports (3020, 3021, 3022) for new backend services. [[1]](diffhunk://#diff-329103b2fffb98e7b96458289b34c5a132f3cbe5b7b2a4b6af71c2f3ae5fee4aL31-R31) [[2]](diffhunk://#diff-fe6aa8dd4ae1c62d32ab1bd7fae58a5cb93f5bc30dc8572b7a1644c25e0aae79R69-R77)

**Test Endpoint Improvements:**

* Improved the `/sec` endpoint in `jaxrs_basic` to return an explicit marker string when the `Authorization` header is missing, making test assertions more robust and clear.

These changes together enhance the reliability of integration test coverage reporting and provide more flexible and precise backend services for testing various gateway behaviors.


all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/features/gateway/endpoint_certificate_invocation.feature
```diff
@@ -0,0 +1,61 @@
...
+    Then The response status code should be 200
+    When I invoke the API at gateway context "{{tlsCertApiContext}}/1.0.0/customers/123/" with method "GET" using access token "generatedAccessToken" and payload "" until response status code becomes 500 within 240 seconds
+    Then The response status code should be 500
+    # ...and it STAYS 500: enforcement, not a transient that the until-status poll happened to catch.
+    When I invoke the API at gateway context "{{tlsCertApiContext}}/1.0.0/customers/123/" with method "GET" using access token "generatedAccessToken" and the response status code should remain 500 for 10 seconds
+
+    Examples:
+      | actor |
+      | admin |
+      | admin@tenant1.com |
```
AnuGayan, line 62: 🔴 **Blocker — this row cannot pass on the current build: the tenant's endpoint certificate never reaches the Gateway Trust Store.**

The `admin@tenant1.com` row fails in CI at line 45 (the LEG 2 poll), after burning the full 240 s:

```
API did not return 200 within the deadline; last response:
{"code":"303001","message":"Runtime Error","description":"Currently, HTTP endpoint :
 [ Name : APIMTlsEndpoint_918b700d_6--v1.0.0_APIproductionEndpoint ] [ State : SUSPENDED ]"}
```

The gateway logs show why, and it is **not** propagation latency — the certificate is simply never published to the gateway:

| time | super-tenant row (`tlsCertAlias_918b700d_1`) | tenant1.com row (`tlsCertAlias_918b700d_5`) |
|---|---|---|
| upload | `01:53:07` added to **Publisher** client Trust Store | `01:55:25` added to **Publisher** client Trust Store |
| propagate | `01:53:08` added to the **Gateway Trust Store** ✅ | **never — no such line for `_5` in the whole run** ❌ |
| result | 200 after the `01:54:03` SSL reload | `SSLHandshakeException: (certificate_unknown) PKIX path building failed` until the deadline |

The HTTPS sender reloads on a 60 s cycle (`01:53:03` / `01:54:03` / `01:55:03`), so 240 s was ample — the certificate genuinely is not in the gateway's trust store for the tenant upload.

Worth noting this is specific to the **endpoint**-certificate path, not to tenant uploads in general: the mTLS matrix scenarios in the same run do propagate their tenant aliases (`PRODUCTION_mtlsOnlyCertAliastenant1com...` reaches the Gateway Trust Store). That makes it look like a real product gap rather than a flaky test.

**How to verify**

```bash
gh api repos/wso2/product-apim/actions/jobs/95556693223/logs > joblog.txt

# every gateway-trust-store event in the run — `_1` is there, `_5` is not:
grep -oE "Alias '[^']*' is successfully (added to|removed from) the Gateway Trust Store" joblog.txt \
  | sort | uniq -c | grep tlsCertAlias

# the tenant alias only ever reaches the PUBLISHER trust store:
grep "tlsCertAlias_918b700d_5" joblog.txt | sed 's/.*: \[2026/[2026/'
```

**Suggested resolution:** drop the `| admin@tenant1.com |` row (keep the super-tenant one, which passes reliably) and raise the tenant propagation gap as a product issue — or confirm the certificate-management event path for tenant-scoped endpoint certificates before this row can carry weight.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/artifacts/payloads/create_apim_tls_endpoint_api.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "${UNIQUE:APIMTlsEndpoint}",
+  "context": "${UNIQUE:apiTlsEndpointContext}",
+  "version": "1.0.0",
+  "endpointConfig": {
+    "endpoint_type": "http",
+    "production_endpoints": {
+      "template_not_supported": false,
+      "config": null,
```
AnuGayan: 🟠 **Endpoint suspension throttles the LEG 2 poll to one usable probe per 30 s.**

With `"config": null` the endpoint inherits the default Synapse suspension. Each failed probe (and LEG 1 guarantees several) suspends the endpoint for a flat 30 000 ms, so the 240 s window in `endpoint_certificate_invocation.feature:45` yields only ~8 real attempts instead of ~120.

This is not the cause of the current failure (see my comment on the feature's Examples row — the certificate never reaches the gateway trust store at all), but it removes most of the headroom the leg needs and will make this scenario flaky even after that is fixed. The failing row logged 8 suspensions; the passing row logged 3.

Pin suspension off in both endpoint blocks:

```json
"config": {
  "actionSelect": "fault", "actionDuration": "30000",
  "factor": "1", "retryDelay": "0", "retryErroCode": [],
  "suspendDuration": "1", "suspendMaxDuration": "1"
}
```

**How to verify**

```bash
# flat 30s suspensions, 8 of them on the failing endpoint:
grep -E "Suspending endpoint : APIMTlsEndpoint" joblog.txt | sed 's/.*Suspending/Suspending/' | cut -c1-200
grep -oE "Suspending endpoint : APIMTlsEndpoint_918b700d_[0-9]+" joblog.txt | sort | uniq -c
```

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/testng-v2.xml
```diff
@@ -547,6 +614,13 @@
             <class name="org.wso2.am.integration.cucumbertests.runners.block.GatewayDenyEnforcementEmailUserNameRunner"/>
             <class name="org.wso2.am.integration.cucumbertests.runners.block.KeyManagerEmailUserNameLoginRunner"/>
             <class name="org.wso2.am.integration.cucumbertests.runners.block.PublisherApiProductEmailUserNameRunner"/>
+            <class name="org.wso2.am.integration.cucumbertests.runners.block.DevPortalCommentsRunner"/>
+            <class name="org.wso2.am.integration.cucumbertests.runners.block.DevPortalRatingsRunner"/>
+            <class name="org.wso2.am.integration.cucumbertests.runners.block.DevPortalVisibilityRunner"/>
+            <class name="org.wso2.am.integration.cucumbertests.runners.block.DevPortalVisibilityDomainsRunner"/>
+            <class name="org.wso2.am.integration.cucumbertests.runners.block.DevPortalSearchRunner"/>
+            <class name="org.wso2.am.integration.cucumbertests.runners.block.PublisherAccessControlRunner"/>
+            <class name="org.wso2.am.integration.cucumbertests.runners.block.KeyManagerTokenIssuanceRunner"/>
```
AnuGayan, line 613: 🔴 **Blocker — registering this runner in the email-username block breaks a pre-existing scenario.**

`token_issuance.feature:342-343` asserts the consent-page application name against a hardcoded `spOwner` column:

```gherkin
And The stored value "consentApplicationName" should be "<spOwner>_{{createdAppId}}_PRODUCTION"
Examples:
  | actor             | spOwner |
  | admin             | admin   |
  | admin@tenant1.com | admin   |
```

Under `emailUserMode` the physical username becomes `admin@email.com`, which the product renders into the service-provider name with `@` → `-AT-`. CI fails exactly there:

```
Stored value 'consentApplicationName' is not the expected value
expected [admin_5eb19d75-..._PRODUCTION] but found [admin-AT-email.com_5eb19d75-..._PRODUCTION]
   at ✽.The stored value "consentApplicationName" should be "admin_{{createdAppId}}_PRODUCTION"
      (token_issuance.feature:342)
```

The scenario cannot pass in this block as written.

**Suggested resolution — either:**
1. remove `KeyManagerTokenIssuanceRunner` from `IntegrationV2-EmailUserName`. Per the block's own selection rule ("listed here only where the PHYSICAL username is load-bearing for an assertion it already makes"), the auth-code consent-name assertion is load-bearing *against* you here — it hardcodes the plain form; or
2. derive `spOwner` at runtime. The PR already adds `I store the username of actor {string} as {string}`, but it returns `admin@email.com` — the SP naming applies the `-AT-` substitution, so that step (or a new one) would need to mirror it.

**How to verify**

```bash
gh api repos/wso2/product-apim/actions/jobs/95586584767/logs | grep -A3 consentApplicationName
```

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/KeyManagerAdminSteps.java
```diff
@@ -115,7 +117,15 @@ public void iChangeTheOwnerOfApplication(String appIdRef, String newOwnerRef) th
         String newOwner = apiUsername(Identity.resolveActor(newOwnerRef));
         Map<String, String> headers = new HashMap<>();
         headers.put(Constants.REQUEST_HEADERS.AUTHORIZATION, "Bearer " + Identity.adminToken());
-        Requests.post(Utils.getChangeApplicationOwnerURL(Utils.getBaseUrl(), appId, newOwner), headers, "", null);
+        HttpResponse response = Requests.post(Utils.getChangeApplicationOwnerURL(Utils.getBaseUrl(), appId, newOwner),
+                headers, "", null);
+        // On a SUCCESSFUL transfer the application now belongs to the new owner, so teardown must delete it AS the
+        // new owner (CLAUDE.md §5): the DevPortal delete is scoped to the requesting subscriber, so sweeping it with
+        // the original creator's token would 404 and leak the application while looking like an already-gone id.
+        if (response.getResponseCode() == 200) {
```
AnuGayan: 🟠 **Unguarded dereference, and a gate that disagrees with its siblings.**

Two things here:

1. `response` is dereferenced without a null check, unlike every other deregister site added in this PR — `ApiCategorySteps` (`response != null && ...`), `MCPServerSteps`, `PublisherBaseSteps.iDeleteTheResource`, `EndpointCertificateSteps.iDeleteEndpointCertificate` all guard first. If `Requests.post` ever yields null here, the step dies with an NPE instead of leaving the application queued for teardown, which is the safe direction.
2. This gates on `== 200` while the siblings gate on the 2xx range. Change-owner answering `204` would silently skip both the deregister and the re-register, and the application would then be swept with the *old* owner's token — the exact leak this block exists to prevent.

```java
if (response != null && response.getResponseCode() >= 200 && response.getResponseCode() < 300) {
    ResourceCleanup.deregister(Constants.CREATED_APPLICATION_IDS, appId);
    ResourceCleanup.registerFor(Constants.CREATED_APPLICATION_IDS, appId, newOwnerRef);
}
```

**How to verify:** `grep -n "getResponseCode() >= 200 && " ` across the step definitions — every other 2026 deregister site in this diff uses the guarded 2xx form.

all-in-one-apim/modules/integration-v2/tests-common/testcontainers/src/main/resources/nodeapps/custom-status-backend/server.js
```diff
@@ -0,0 +1,65 @@
...
+// socket literally. Everything else about the response is well formed (accurate Content-Length, explicit
+// Connection: close) so the only unusual thing on the wire is the reason phrase under test.
+//
+// EVERY request on this port gets the custom response, whatever the path — the port is the selector, exactly
+// as in the legacy socket server. `/custom-status` is the path tests should use so the intent reads clearly in
+// the feature, but nothing here depends on it.
+
+const net = require('net');
+
+const port = 3024;
```
AnuGayan: 🟠 **The port PM2 assigns is ignored — `env.PORT` here is dead configuration.**

`ecosystem.config.js` sets `env: { PORT: 3024 }` for this app, but the server hardcodes `const port = 3024`. The same applies to `tls-backend/server.js:40` (`PORT: 3023`).

The value now has to agree in four places — `ecosystem.config.js`, this constant, `NodeAppServer.exposedPorts`, and the Dockerfile `EXPOSE`. Because `Wait.forListeningPort` waits on **all** exposed ports, a future edit to only the ecosystem entry leaves the server on the old port and every block using this image stalls at boot with no indication why.

```js
const port = Number(process.env.PORT) || 3024;
```

**How to verify**

```bash
# the declared env var is never read by either new backend:
grep -rn "process.env.PORT" nodeapps/custom-status-backend nodeapps/tls-backend   # -> no hits
grep -n "PORT" nodeapps/ecosystem.config.js | tail -4                             # -> 3023 / 3024 declared
```

all-in-one-apim/modules/integration-v2/tests-common/testcontainers/src/main/resources/nodeapps/custom-status-backend/server.js
```diff
@@ -0,0 +1,65 @@
...
+const server = net.createServer((socket) => {
+  let received = '';
+  socket.on('data', (chunk) => {
+    received += chunk.toString();
+    // Answer as soon as the request head is complete; the canned response is the same regardless of the
+    // request, so there is no reason to read a body.
+    if (received.includes('\r\n\r\n')) {
+      console.log('----invoking custom-status backend, request line: ' + received.split('\r\n')[0]);
+      socket.end(response);
+    }
```
AnuGayan, line 72: 🟠 **`received` grows without bound, and a second `data` chunk re-enters `socket.end`.**

Two issues in this handler:

* a client that never sends `\r\n\r\n` keeps appending to `received` for the lifetime of the connection — unbounded memory for anything that opens the socket and holds it;
* after `socket.end(response)`, any further chunk still satisfies `received.includes('\r\n\r\n')`, so `socket.end()` is called again. The resulting write-after-end error is swallowed by the ignore-everything `error` handler below, so it fails silently rather than loudly.

Blast radius is small (a test container), but the fix is three lines:

```js
let answered = false;
socket.on('data', (chunk) => {
  if (answered) return;
  received += chunk.toString();
  if (received.length > 64 * 1024) { socket.destroy(); return; }
  if (received.includes('\r\n\r\n')) {
    answered = true;
    console.log('----invoking custom-status backend, request line: ' + received.split('\r\n')[0]);
    socket.end(response);
  }
});
```

**How to verify** (against the running container, `nodebackend:3024`):

```bash
# double-end: send the head, then more bytes on the same socket
printf 'GET /custom-status HTTP/1.1\r\nHost: x\r\n\r\nEXTRA' | nc nodebackend 3024
# unbounded: a head that never terminates keeps the buffer growing
python3 -c "import socket,time;s=socket.create_connection(('nodebackend',3024));[s.send(b'A'*65536) or time.sleep(0.1) for _ in range(100)]"
```

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/EndpointCertificateSteps.java
```diff
@@ -202,14 +224,78 @@ public void theSearchShouldReturnNCertificates(int expected) {
...
+                "Certificate validity 'from' mismatch; body: " + validity);
+        Assert.assertEquals(validity.optString("to", null), expectedTo,
+                "Certificate validity 'to' mismatch; body: " + validity);
+    }
+
+    /**
+     * The last published response's body as JSON, guarded so a failed/empty response fails with the status and body
+     * rather than an opaque JSONException from the parse (§7).
+     */
+    /**
```
AnuGayan, line 321: 🟡 **Two consecutive javadoc blocks — the first one is orphaned.**

Lines 288-291 and 292-296 are both javadoc for `lastResponseBody()`. Only the second attaches; the first becomes a stray comment that says something the code no longer does (it describes a local guard, while the method now delegates to `Utils.requireJsonBody`). Leftover from the rewrite — delete lines 288-291.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/KeyManagerAdminSteps.java
```diff
@@ -184,6 +215,81 @@ public void theStoredValueShouldEqual(String keyA, String keyB) {
...
+     * AND that its {@code additionalProperties.description} equals {@code expectedDescription} exactly — i.e. the
+     * description supplied at generation round-trips. Needs its own step because it is a lookup of one list entry BY
+     * ID followed by a nested-property comparison; the JSONPath field step cannot express it (it resolves
+     * placeholders only in the expected value, not inside the path).
+     */
+    @Then("The listed consumer secret {string} should have description {string}")
+    public void theListedSecretShouldHaveDescription(String secretIdKey, String expectedDescription) {
+        String secretId = TestContext.resolve(secretIdKey).toString();
+        JSONObject body = successBody("consumer secrets list");
+        org.json.JSONArray list = body.getJSONArray("list");
```
AnuGayan: 🔵 **Fully-qualified `org.json.JSONArray` while `JSONObject` is imported.**

`JSONObject` is imported at the top of this file; `JSONArray` is spelled out inline only here. Add the import and use the simple name, matching every other step class in the package (`EndpointCertificateSteps`, `GatewayRestArtifactsSteps`, and `PublisherBaseSteps` all import it).

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/BaseSteps.java
```diff
@@ -1309,17 +1411,68 @@ public void theResourceShouldReflectTheUpdatedAsValueFromContext(String resource
...
      * @param config The configuration field name to check
      * @param configValue The expected configuration value
      */
-    private void verifyConfigurationInResponse(HttpResponse response, String config, String configValue) {
-        Assert.assertTrue(response != null && response.getData() != null && !response.getData().isBlank(),
-                "No response with a body to verify configuration '" + config + "' in; got="
-                        + (response == null ? "null" : response.getResponseCode() + "/" + response.getData()));
+    private void verifyConfigurationInResponse(HttpResponse response, String config, String configValue,
+                boolean unorderedArray) {
+        Assert.assertTrue(response != null && response.getData() != null && !response.getData().isEmpty(),
```
AnuGayan: 🟡 **Guard weakened from `isBlank()` to `isEmpty()`, which re-opens the failure mode it exists to close.**

A whitespace-only body now passes this guard and reaches `new JSONObject(response.getData())` on the next line, throwing an opaque `JSONException` — precisely the outcome the guard was added to prevent (and the one the sibling guards in this same file still prevent with `isBlank()`).

The message also lost the body it used to print, which is the half a reader actually needs when this fires:

```java
Assert.assertTrue(response != null && response.getData() != null && !response.getData().isBlank(),
        "No response with a body to verify configuration '" + config + "' in; got="
                + (response == null ? "null" : response.getResponseCode() + "/" + response.getData()));
```

**How to verify:** `grep -n "isBlank()" BaseSteps.java` — every other response guard in the file uses `isBlank`; this is now the only `isEmpty`.

AnuGayan (commented):
## Review — Group 3 Parity

Reviewed the full diff (139 files, ~9.9k additions): the new runners and feature suites, the Java step definitions, `render_coverage_tree.py`, the two new node backends, and the TestNG wiring. I also pulled the failing CI job's log and test report to attribute the red gate.

**Overall this is strong work** — the execution-verified coverage tree is a real improvement (a tagged-but-never-run scenario counting as coverage is worse than an admitted gap), the cleanup deregistration fixes close genuine leaks, and the `iShouldBeAbleToRetrieveSubscription` fix corrects a bug that could hand a caller *another application's* subscription id. The rationale comments throughout are unusually good.

**Verdict: 🔧 request changes** — `integration-v2 (v2-1, testng-v2.xml)` is red (`5971 tests run, 4 skipped, 6 failed`; 3 unique failures), and two are attributable to this PR.

### The three CI failures

| # | Failure | Attribution |
|---|---|---|
| 1 | `GatewayEndpointCertificateInvocationRunner` — the `admin@tenant1.com` row times out at LEG 2; the certificate reaches the Publisher trust store but **never the Gateway Trust Store** | new in this PR — see inline comment on the feature's Examples row |
| 2 | `KeyManagerTokenIssuanceRunner` — `consentApplicationName` expects `admin_…`, gets `admin-AT-email.com_…` | new in this PR — caused by adding the runner to the email-username block |
| 3 | `AdminThrottlingPolicyRunner` — product `500` on `PUT /apis/{id}` updating `policies` (`admin/throttling_policy.feature:239`) | **not** attributable to this PR — that feature and runner are untouched |

On (3): I checked the last fully-green `integration-v2` job (run `31926192028`, branch `integration-v2-group1-port`, 2026-08-16) and this scenario is not among its failures, so it is not a chronic red. That leaves "intermittent under the Admin block's parallel load" vs "perturbed by the added load" — a re-run is the cheapest way to classify it. If it does not reproduce, worth filing separately rather than holding this PR.

Also confirmed while reviewing: all 10 new runners are registered in `testng-v2.xml`, the 5 usages of the removed `The service catalog search should return {int} entries` step are all removed with it, and the full log contains no undefined steps and no assertion errors beyond the three above.

Inline comments below: 2 blockers, 4 major, 5 minor/nits. Each carries a way to reproduce or check it.

<sub>Note: `tls-backend/certs/server.key` is a committed PEM private key. It is a self-signed, test-only fixture and the README justifies committing it well — but Gitleaks/secret scanning will flag it, so it is worth allowlisting the path so this does not become recurring review noise.</sub>

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/features/gateway/basic_auth_security.feature
```diff
@@ -0,0 +1,226 @@
...
+
+  # Ports testInvokeInternalKeyForBasicAuthOnlyAPI — a deliberate POSITIVE, not a negative. The publisher's
+  # internal API key BYPASSES the API's declared application-security scheme: it invokes a basic_auth-ONLY API
+  # successfully (200), even though the OAuth token and API key above are both refused on the same API. That is
+  # the internal key's purpose (the publisher try-out path, which must work whatever the API declares), and
+  # asserting it keeps a future change that "fixed" the internal key into a 401 from passing silently.
+  #
+  # Runs x2 tenants so the internal-key scheme bypass is verified in each tenant.
+
+    Examples:
```
AnuGayan: 🟡 **This `Examples` table belongs to the scenario ~50 lines above, but sits below the *next* scenario's comment block.**

Lines 178-189 document "An internal API key invokes a basic_auth-only API…" (which starts at line 197), yet they are placed between the previous scenario's last step and its `Examples` table here. Gherkin ignores comments anywhere, so it parses and runs correctly — but read top-to-bottom it looks like this table drives the internal-key scenario, which is the sort of thing that gets "fixed" wrongly later.

Move the comment block down so it sits immediately above the `@cap:gateway …` tag line at 196.

all-in-one-apim/modules/integration-v2/docs/devs/render_coverage_tree.py
```diff
@@ -8,25 +8,47 @@
...
 
 EXCLUSION = {"infra", "framework", "migration", "setup"}
 SETUP_PREFIX = "_setup_"
 VALID_TYPES = {"smoke", "negative", "regression"}
 SCENARIO_RE = re.compile(r"^\s*(Scenario Outline|Scenario):\s*(.*)$")
 FEATURE_RE = re.compile(r"^\s*Feature:\s*(.*)$")
+CLASS_RE = re.compile(r'<class\s+name="([^"]+)"')
+XML_COMMENT_RE = re.compile(r"<!--.*?-->", re.S)
+JAVA_BLOCK_COMMENT_RE = re.compile(r"/\*.*?\*/", re.S)
+JAVA_LINE_COMMENT_RE = re.compile(r"//[^\n]*")
```
AnuGayan: 🟡 **`//` inside a string literal is stripped as a comment.**

`JAVA_LINE_COMMENT_RE = re.compile(r"//[^\n]*")` is applied to the whole file before the `features = …` match. A runner with a URL in a string on the same line as the options (`features = "…"`, anything containing `http://`) loses the rest of that line, `FEATURES_OPT_RE` then finds nothing, and the runner maps to `[]` features — which this tool reports as **NOT EXECUTED**, exactly the false signal the check was built to eliminate. The failure is silent and looks like a real coverage gap.

No runner trips it today, so this is a latent trap rather than a live bug. Worth at least a comment; if you want it robust, strip line comments only outside quoted strings, or match `features` before stripping.

**How to verify**

```bash
cd all-in-one-apim/modules/integration-v2/docs/devs
python3 - <<'EOF'
import re
JAVA_LINE_COMMENT_RE = re.compile(r"//[^\n]*")
src = 'features = "src/test/resources/features/x.feature", // see http://example.com\nglue = {}'
print(repr(JAVA_LINE_COMMENT_RE.sub("", src)))   # the features value survives here...
src2 = '@CucumberOptions(features = "http://x/a.feature")'
print(repr(JAVA_LINE_COMMENT_RE.sub("", src2)))  # ...but any // inside the value truncates it
EOF
```

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/runners/block/GatewayResourceAuthTypeRunner.java
```diff
@@ -0,0 +1,41 @@
...
+@CucumberOptions(
+        features = {
+                "src/test/resources/features/gateway/resource_auth_type.feature"
+        },
+        glue = {
+                "org.wso2.am.integration.cucumbertests.stepdefinitions"
+        },
+        plugin = {"pretty", "html:target/cucumber-report/gateway-resource-auth-type.html"}
+)
+public class GatewayResourceAuthTypeRunner extends BaseBlockRunner {
```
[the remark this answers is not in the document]
nimsara66: @coderabbitai,

please resolve if not applicable anymore

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/utils/TenantUserProvisioner.java
```diff
@@ -585,6 +645,41 @@ public static void deleteUser(String tenantDomain, String userName) throws IOExc
...
+                + "xmlns:ser=\"http://service.ws.um.carbon.wso2.org\">"
+                + "<soapenv:Header/><soapenv:Body>"
+                + "<ser:updateCredentialByAdmin>"
+                + "<ser:userName>" + Utils.escapeXml(userName) + "</ser:userName>"
+                + "<ser:newCredential>" + Utils.escapeXml(newPassword) + "</ser:newCredential>"
+                + "</ser:updateCredentialByAdmin>"
+                + "</soapenv:Body></soapenv:Envelope>";
+        HttpResponse response = SimpleHTTPClient.getInstance().sendSoapRequest(
+                Utils.getRemoteUserStoreManagerServiceURL(Utils.getBaseUrl()), payload,
+                "urn:updateCredentialByAdmin", tenantAdmin.getUserName(), tenantAdmin.getPassword());
```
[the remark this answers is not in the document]
nimsara66: @coderabbitai,

please resolve if not applicable anymore

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/features/key-manager/token_revocation.feature
```diff
@@ -43,3 +44,78 @@ Feature: Key Manager Token Revocation
...
+  # issued under the OLD one is still honoured at the gateway.
+  #
+  # The resource owner is a principal this scenario provisions at runtime, NOT one of the block's seeded actors:
+  # those live in the block-shared actor registry with their passwords cached there, so resetting one would break
+  # every sibling runner in the block. Because that principal is deliberately not registered as an actor, the
+  # token is requested with the EXPLICIT-credentials password-grant step.
+  #
+  # Runs x2 tenants so token invalidation on credential change is verified in each tenant's key-manager context.
+  @cap:key-manager @feat:token-revocation @rule:password-change @type:regression @dep:gateway @dep:admin @legacy:APISecurityTestCase
+  Scenario Outline: An already-issued access token is rejected after its user's password is reset
```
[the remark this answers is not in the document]
nimsara66: @coderabbitai,

please resolve if not applicable anymore

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/features/publisher/service_catalog.feature
```diff
@@ -160,3 +168,170 @@ Feature: Publisher Service Catalog
...
+  # archive; the step proves it is a real, complete archive (non-empty, unzips, carries both metadata and
+  # definition) rather than merely a 200. A wrong name with an existing version is 404.
+  @cap:publisher @feat:service-catalog @type:regression @rule:export @legacy:ServiceCatalogRestAPITestCase
+  Scenario Outline: Export a service catalog entry as <actor>
+    Given The system is ready
+    And I have valid access tokens as "<actor>"
+    And I generate a unique value and store it as "expKey"
+    And I generate a unique value and store it as "expName"
+    When I create a service catalog entry named "{{expName}}" version "v1" key "{{expKey}}" from definition "artifacts/service-catalog/definition1.yaml" as "expSvcId"
+    Then The response status code should be 200
```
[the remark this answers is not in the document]
nimsara66: @coderabbitai,

please resolve if not applicable anymore

all-in-one-apim/modules/integration-v2/docs/devs/group3-parity-findings.md
```diff
@@ -0,0 +1,626 @@
...
+parsed moments earlier suggests a race or shared parser state, not bad input.
+
+### F4 — The array assertion's "order-insensitive" message was false *(fixed in this work)*
+
+`BaseSteps.verifyConfigurationInResponse` compared arrays with `JSONArray.similar()`, which walks INDEX BY INDEX
+and is order-**sensitive**, while telling the reader the opposite:
+
+```
+Expected (order-insensitive): ["Unlimited","Gold"]  But got: ["Gold","Unlimited"]
+```
```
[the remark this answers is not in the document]
nimsara66: @coderabbitai,

please resolve if not applicable anymore

all-in-one-apim/modules/integration-v2/docs/devs/render_coverage_tree.py
```diff
@@ -76,6 +98,67 @@ def parse_tags(line):
...
+    if registered is None:
+        return None, None
+    runner_features = parse_runner_features(runners_dir)
+    if not runner_features:
+        return None, None
+    executed = set()
+    for cls in registered:
+        for f in runner_features.get(cls, []):
+            executed.add(feature_rel(f))
+    return executed, registered
```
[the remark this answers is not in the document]
nimsara66: @coderabbitai,

please resolve if not applicable anymore

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/runners/block/GatewayResourceAuthTypeRunner.java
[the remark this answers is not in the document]
nimsara66: @coderabbitai resolve

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/SubscriptionListingSteps.java
```diff
@@ -51,10 +51,32 @@ public class SubscriptionListingSteps {
...
+     * fails with the real body rather than being hidden by the poll.
+     */
+    @When("I retrieve all subscriptions of api {string} until the list contains {int} subscriptions within {int} seconds")
+    public void iRetrieveAllSubscriptionsOfApiUntilCount(String apiIdKey, int expectedCount, int timeoutSeconds)
+            throws InterruptedException {
+        Utils.retryUntil(timeoutSeconds * 1000L,
+                () -> listSubscriptionsOfApi(apiIdKey),
+                response -> response != null && response.getResponseCode() == 200 && response.getData() != null
+                        && new JSONObject(response.getData()).getJSONArray("list").length() == expectedCount);
+    }
```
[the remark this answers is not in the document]
nimsara66: @coderabbitai,

Both points are already in place: the predicate guards response == null || != 200 || getData() == null || isBlank() before new JSONObject(...) (lines 73–76). The probe calls Requests.get, which publishes each attempt to httpResponse, so the final poll's response is already what the following assertion reads — assigning retryUntil's return would re-publish the same object. No change made.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/utils/TenantUserProvisioner.java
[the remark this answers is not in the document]
nimsara66: @coderabbitai resolve

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/features/key-manager/token_revocation.feature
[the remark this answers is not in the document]
nimsara66: @coderabbitai resolve

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/features/key-manager/token_issuance.feature
```diff
@@ -122,12 +130,198 @@ Feature: Key Manager Token Issuance
...
+    When I retrieve all APIs created through the Publisher REST API
+    Then The response status code should be 200
+
+    # Devportal plane, symmetrically: a creator's token cannot list applications; a subscriber's can.
+    Given The system is ready and I have valid devportal access token as "roleScopeCreator<suffix>"
+    When I retrieve all applications from the Developer Portal
+    Then The response status code should be 401
+    Given The system is ready and I have valid devportal access token as "subscriberUser<suffix>"
+    When I retrieve all applications from the Developer Portal
+    Then The response status code should be 200
```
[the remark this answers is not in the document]
nimsara66: @coderabbitai,

Lines 276/279 (not 312–318, which are comments) already resolve — The system is ready and I have valid devportal access token as {string} is its own registered step, distinct from I have a valid Devportal access token as {string}. The composite additionally waits for readiness, which these scenarios rely on as they have no separate Given The system is ready; switching phrasing would drop that. No change made.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/features/gateway/endpoint_certificate_invocation.feature
nimsara66: Run 2 of the same suite is green (1954/0) and its log shows tlsCertAlias_fbea787d_5_1 both added to and removed from the Gateway Trust Store — the tenant endpoint-certificate path works; addCertificateToGateway appends _<tenantId> for non-super tenants (CertificateManagerImpl:200-207), which is why a grep for the bare _5 alias comes up empty.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/artifacts/payloads/create_apim_tls_endpoint_api.json
nimsara66: Valid, fixed on both payloads (create_apim_tls_endpoint_api.json, create_apim_jsontoxml_api.json). One correction: suspendDuration must be "0", not "1" — TemplateBuilderUtil disables suspension only on an exact == 0; "1" falls through and leaves it wired.

Measured margin in the passing run was ~2s: cert landed 16:01:29, a probe at 16:01:58 raced 9s ahead of the 16:02:07 SSL reload, re-suspended 30s, LEG 3 started 16:02:30.

Scoped empirically (68 payloads have config: null — it's the Publisher default): only these two endpoints ever suspend. Note the JsonToXml one is correctness, not headroom — suspension replaces the asserted 601000 with 303001.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/testng-v2.xml
nimsara66: Stale — already fixed in 057eda345 (an earlier feedback round). The <spOwner> Examples column no longer exists; the assertion reads {{spOwnerName}}, derived at runtime by ApplicationBaseSteps.serviceProviderOwnerName().

That derivation handles both axes: it strips the tenant suffix (a tenant1.com admin still owns SPs as plain admin) and applies @ → -AT-, so it yields admin normally and admin-AT-email.com under emailUserMode — exactly your option 2.

Job 95586584767 predates the fix. KeyManagerTokenIssuanceRunner stays in the block; fv7 will confirm it at runtime.

all-in-one-apim/modules/integration-v2/tests-common/testcontainers/src/main/resources/nodeapps/custom-status-backend/server.js
nimsara66: Stale — already fixed in 057eda345, identical to your patch: an answered flag guarding re-entry into socket.end, and a MAX_HEAD_BYTES = 64 * 1024 cap that socket.destroy()s a client that never terminates its head.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/EndpointCertificateSteps.java
nimsara66: Stale — the orphaned block existed at f3d02ee00 and was deleted in 057eda345. lastResponseBody() now carries exactly one javadoc, and it correctly describes the Utils.requireJsonBody delegation rather than a local guard.

all-in-one-apim/modules/integration-v2/docs/devs/render_coverage_tree.py
nimsara66: Valid and latent — confirmed the truncation reproduces (features = "http://x/a.feature" → option unmatched → [] → NOT EXECUTED) and that no runner trips it today. Replaced the regex with a hand-rolled strip_java_comments scan that skips string/char literals and honours escapes.

Also made the failure loud instead of silent: a concrete runner with no parseable features option now warns on stderr (the abstract BaseBlockRunner is exempt — it carries no @CucumberOptions by design, which the warning itself surfaced on first run).

Validated: 7/7 unit cases (URL-in-value, trailing // comment, braced array, block comment, escaped quote, char literal, decoy in a comment), regenerated coverage-tree.md is byte-identical, and a control confirms the warning fires for a genuine parse failure and stays quiet for the base class.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/features/publisher/service_catalog.feature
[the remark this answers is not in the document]
nimsara66: @coderabbitai,

The test currently asserts the existing product behaviour, and the inconsistency has been reported as a bug in https://github.com/wso2/api-manager/issues/5195. 

Therefore, fixing the product inconsistency is not a concern for this PR.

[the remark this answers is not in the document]
nimsara66: @coderabbitai resolve

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/features/gateway/audience_validation.feature
```diff
@@ -52,6 +54,93 @@ Feature: Gateway Audience Validation
...
+    And I extract response field "token.accessToken" and store it as "audProdSandboxToken"
+    When I put the following JSON payload in context as "audProdSubPayload"
+    """
+    {"applicationId": "{{applicationId}}", "apiId": "{{apiId}}", "throttlingPolicy": "Unlimited"}
+    """
+    And I subscribe to API "audProductId" using application "createdAppId" with payload "audProdSubPayload" as "audProdSubId"
+    Then The response status code should be 201
+
+    # No audiences configured on the product -> the sandbox token is accepted (200).
+    When I invoke the API at gateway context "{{audProductContext}}/1.0.0/customers/123/" with method "GET" using access token "audProdSandboxToken" and payload "" until response status code becomes 200 within 60 seconds
```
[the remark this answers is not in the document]
nimsara66: @coderabbitai,

Skipping — the token must keep the SANDBOX consumer key as its aud, since line 135 configures audiences as ["Hello","{{consumerKey}}"] and line 140 expects access restored. Re-minting from a different path risks a different aud and breaks that leg; aud is fixed at issue time, so re-minting tests nothing new.

The ordering isn't a defect either: a pre-subscription token would fail at line 112 with 900908/403, but a live run exercised all 4 outline rows with audProdSandboxToken 8 times — 0 failures, 0 900908. The gateway validates the subscription against its store at invocation, not from issue-time claims.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/features/gateway/rest_invocation.feature
```diff
@@ -746,6 +763,101 @@ Feature: Gateway REST API Invocation
     When I invoke the API at gateway context "{{lbContext}}/1.0.0/name" with method "GET" using access token "generatedAccessToken" and payload "" until response body contains "File 3" within 60 seconds
     Then The response status code should be 200
 
+    # ORDER: over nine consecutive calls every window of three hits three DISTINCT members — real round-robin, not
+    # merely "all three are reachable".
+    When I invoke the API at gateway context "{{lbContext}}/1.0.0/name" with method "GET" using access token "generatedAccessToken" 9 times round-robin across backend markers "File 1,File 2,File 3"
```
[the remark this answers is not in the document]
nimsara66: @coderabbitai,

Skipping — already prevented. APIInvocationSteps.java:294-300 asserts each response matches exactly ONE marker before the window logic runs, so an overlapping set fails loudly rather than satisfying a window falsely; observed[] holds resolved markers, so no substring semantics reach the distinctness check.

Both call sites have non-overlapping markers today (File 1,File 2,File 3, File 1_Sandbox,File 2_Sandbox,File 3_Sandbox), verified programmatically.

Boundary matching would also weaken it: \bFile 1\b wouldn't match File 1_Sandbox (_ is a word char), so an ambiguous set would resolve silently instead of failing as the test bug it is. Exact call count and sliding-window checks are unchanged either way.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/SubscriptionListingSteps.java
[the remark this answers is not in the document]
nimsara66: @coderabbitai,

I have addressed the concern. Can you re-review?

nimsara66:
@coderabbitai,

Please approve

AnuGayan (approved):

PR #14289, opened 2026-08-10
Add Websocket Proxy Test Cases

hisanhunais:
### Purpose

Adds end-to-end integration tests for the [[transport.ws.proxy_profile]] TOML configuration, which routes outbound WebSocket backend connections through an HTTP CONNECT proxy. Tests run against a real Squid HTTP proxy in a Docker container with CONNECT count assertions drawn directly from Squid's access logs.

#### Infrastructure

**Squid proxy container (tests-common/testcontainers)**
- New SquidProxyServer Testcontainers singleton with two Squid instances in one container:
  - Port 3128 — anonymous proxy (no credentials)
  - Port 3129 — Basic-auth proxy (testproxyuser / testproxypass)
- Both instances join ContainerNetwork.SHARED_NETWORK under the alias squid-proxy, reachable from APIM containers without host networking
- SquidProxyServer exposes getAnonConnectCount(), getAuthConnectCount(), and clearLogs() via grep | wc -l against per-instance Squid access logs
- BlockLifecycleListener extended with a PROXY_PERMIT semaphore (mirroring IS_NOTIFY_ALIAS_PERMIT) that serialises proxy blocks JVM-wide while non-proxy blocks run concurrently — prevents log truncation races across the shared Squid container

**Build (pom.xml)**
- squid.docker.image.name system property added to integration-v2/pom.xml and forwarded through both surefire executions
- tests-common/testcontainers/pom.xml — build-squid-proxy-docker-image Maven exec builds the image at pre-integration-test

**Suite files**
- testng-v2.xml — 7 proxy blocks appended to the full suite

#### Tests added

| Feature file | Scenario | What it asserts |
|---|---|---|
| `websocket_proxy_anon_routing` | Proxy profile without bypass — gateway routes through anonymous Squid | Echo succeeds; anon CONNECT count = 1 |
| `websocket_proxy_anon_routing` | Proxy routes multiple WS connections — CONNECT count accumulates | Echo succeeds twice; anon CONNECT count = 2 |
| `websocket_proxy_auth_routing` | Proxy profile with Basic-auth credentials — gateway authenticates with Squid | Echo succeeds; auth CONNECT count = 1 |
| `websocket_proxy_auth_wrong_creds` | Wrong proxy credentials — gateway fails to establish proxy tunnel | WS rejected; anon count = 0; auth count = 0 |
| `websocket_proxy_catch_all_bypass` | `bypass_hosts` in a catch-all profile — gateway connects directly | Echo succeeds; anon CONNECT count = 0 |
| `websocket_proxy_no_match` | Backend hostname unmatched by any profile — gateway connects directly | Echo succeeds; anon CONNECT count = 0 |
| `websocket_proxy_specific_bypass` | `bypass_hosts` in a specific profile — gateway connects directly | Echo succeeds; anon CONNECT count = 0 |
| `websocket_proxy_specific_over_catch_all` | Specific profile beats catch-all regardless of domain | Echo succeeds; anon CONNECT count = 1 |

#### New step definitions (WebSocketProxySteps)

Given the proxy access logs are cleared
Then the anonymous proxy should have received exactly {int} CONNECT request(s)
Then the authenticated proxy should have received exactly {int} CONNECT request(s)

#### Notable design decisions

- CONNECT count over echo alone — Squid can reach nodebackend on the Docker network, so a successful echo does not prove the gateway used the proxy. Zero CONNECT entries are the only reliable signal for bypass and no-match scenarios.
- Catch-all declared first in precedence TOML — makes the precedence test distinguish first-match from most-specific-match semantics. A first-match implementation would select the catch-all and bypass the proxy (CONNECT = 0, fail); a most-specific-match implementation selects the specific profile regar1, pass).
- Wrong-creds asserts both counts = 0 — APIM's WS proxy client fails before sending CONNECT to Squid when credentials are wrong. Asserting both proxy counts = 0 rules out accidental anon-proxy fallback and is more specific than WS rejection alone.


hisanhunais:
@coderabbit review

hisanhunais:
@coderabbitai review

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/artifacts/payloads/create_apim_ws_proxy_api.json
```diff
@@ -0,0 +1,46 @@
...
+  "additionalPropertiesMap": {},
+  "accessControl": "NONE",
+  "endpointConfig": {
+    "endpoint_type": "ws",
+    "sandbox_endpoints": {
+      "url": "ws://proxied.backend.test:3001"
+    },
+    "production_endpoints": {
+      "url": "ws://proxied.backend.test:3001"
+    }
```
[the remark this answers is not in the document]
hisanhunais: Fixed. The root cause was create_apim_ws_proxy_api.json using ws://proxied.backend.test:3001 as the backend, while all proxy profile TOMLs have target_hosts = ["nodebackend"] (or "*" for the catch-all). The payload is now ws://nodebackend:3001, making it consistent with the rest of the suite.

The TOML files themselves did not need changes — their target_hosts values were already correct. wsProxyNoMatch/deployment.toml intentionally uses target_hosts = ["proxied\\.backend\\.test"] as a decoy that does not match nodebackend, which is what gives that block its "no profile matches" semantics.

create_apim_ws_proxy_api.json is also now unreferenced by any active feature file (the one scenario that used it was removed as part of this fix), so the inconsistency cannot recur.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/testng-ws-proxy.xml
```diff
@@ -0,0 +1,78 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
nimsara66: @hisanhunais,
Why do we need a separate suite?

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/testng-v2.xml
```diff
@@ -578,4 +578,81 @@
...
         </classes>
     </test>
+
+    <!-- WebSocket proxy profiles. Each block runs its own APIM container with a distinct TOML overlay
+         so the proxy scenarios can run concurrently. A JVM-wide SquidProxyServer (initProxy=true)
+         provides two Squid instances on the Docker network: anonymous on port 3128 and Basic-auth on
+         port 3129. CONNECT count assertions on the Squid access logs prove gateway routing behaviour;
+         because Squid can reach nodebackend, echo success alone would not prove bypass or no-match.
+         Needs the node WS echo backend (initBackend). Ports WebSocketProxyProfileTestCase. -->
+    <test name="IntegrationV2-WsProxySpecificBypass" parallel="classes" thread-count="1">
```
nimsara66, line 842: Check if these single runner blocks can be consolidated into existing single runner test blocks without contamination

hisanhunais: These blocks can't be consolidated — each tests a structurally distinct proxy profile configuration ([[transport.ws.proxy_profile]]) that is loaded into a JVM-wide singleton at server startup and cannot change between scenarios, so sharing a container would either break the proxy test or silently route the host block's WS connections through Squid. It's also better to keep this behaviour isolated in its own blocks rather than embedding it inside an unrelated test block, since proxy routing is a distinct gateway concern that deserves explicit, dedicated coverage.

nimsara66 (approved):

chamilaadhi (approved):

PR #14299, opened 2026-08-24
Cover Group 4 Parity

nimsara66:
This pull request updates the API test coverage documentation to reflect new and reorganized product scenarios, including additional test cases and improved coverage in several key areas. The changes mainly involve increasing the number of covered scenarios, updating feature references, and expanding test cases for API lifecycle, operation policies, versioning, import/export, streaming, and SOAP API design.

**Coverage Increases and Scenario Additions:**

* The total number of covered product scenarios increased from 833 to 893, reflecting significant test coverage expansion.
* Several branches now include more detailed and granular test cases, especially in API lifecycle, operation policies, versioning, definitions import/export, streaming, and SOAP API design. [[1]](diffhunk://#diff-39b77fe3d81db566bad1b67b2d8849fbac066f56e590f2694542d003d1524a07L24-R35) [[2]](diffhunk://#diff-39b77fe3d81db566bad1b67b2d8849fbac066f56e590f2694542d003d1524a07L67-R85) [[3]](diffhunk://#diff-39b77fe3d81db566bad1b67b2d8849fbac066f56e590f2694542d003d1524a07L91-R101) [[4]](diffhunk://#diff-39b77fe3d81db566bad1b67b2d8849fbac066f56e590f2694542d003d1524a07L111-R133) [[5]](diffhunk://#diff-39b77fe3d81db566bad1b67b2d8849fbac066f56e590f2694542d003d1524a07L181-R217)

**Key Thematic Changes:**

**1. API Lifecycle and Versioning**
- Added new lifecycle scenarios, such as re-creating deleted APIs and more granular state transitions.
- Expanded versioning coverage, including notifications to subscribers and handling of non-semver versions.

**2. Operation Policies**
- Increased operation policy tests from 9 to 11, adding scenarios for JSON-spec and secret-attribute API-specific policies.

**3. Import/Export and Definitions**
- Expanded import/export scenarios to cover lifecycle state preservation, provider ownership, scope round-trips, and CORS configuration.

**4. Streaming and SOAP API Design**
- Added new streaming API scenarios and expanded SOAP design tests from 9 to 11, including multi-file WSDL imports and policy updates for SOAP-to-REST.

**5. Reference and Line Number Updates**
- Updated feature references and line numbers throughout to match the latest test definitions and maintain traceability. [[1]](diffhunk://#diff-39b77fe3d81db566bad1b67b2d8849fbac066f56e590f2694542d003d1524a07L24-R35) [[2]](diffhunk://#diff-39b77fe3d81db566bad1b67b2d8849fbac066f56e590f2694542d003d1524a07L67-R85) [[3]](diffhunk://#diff-39b77fe3d81db566bad1b67b2d8849fbac066f56e590f2694542d003d1524a07L91-R101) [[4]](diffhunk://#diff-39b77fe3d81db566bad1b67b2d8849fbac066f56e590f2694542d003d1524a07L111-R133) [[5]](diffhunk://#diff-39b77fe3d81db566bad1b67b2d8849fbac066f56e590f2694542d003d1524a07L181-R217)


all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/runners/block/WorkflowApprovalRunner.java
```diff
@@ -49,7 +49,11 @@
                 "src/test/resources/features/admin/workflow_application_deletion.feature",
                 "src/test/resources/features/admin/workflow_key_registration.feature",
                 "src/test/resources/features/admin/workflow_pending_cleanup.feature",
-                "src/test/resources/features/admin/workflow_user_signup.feature"
+                "src/test/resources/features/admin/workflow_user_signup.feature",
+                // MUST remain the lexicographically LAST feature here: it flips APIStateChange to the external
+                // BPMN executor for the rest of the run (see the file's own note). Array order is not honoured —
+                // cucumber-testng sorts by filename — so the "ws" prefix is what keeps it last.
+                "src/test/resources/features/admin/workflow_ws_api_state_change.feature"
```
AnuGayan: 🟠 **major — correctness hinges on lexicographic feature-file ordering**

This feature flips `APIStateChange` to `APIStateChangeWSWorkflowExecutor` (a server-global registry write, per tenant) and never flips it back; safety rests entirely on the filename sorting last among the seven `workflow_*.feature` entries here. `WorkflowAdminSteps#iEnableApprovalWorkflowExecutors` captures the original only once per tenant and the restore happens in `@AfterClass`, so any sibling that sorts after `workflow_ws_*` (a rename, or a new `workflow_x*`/`workflow_z*`) silently routes its publish requests at the BPMN double on `nodebackend:3004`. That surfaces as "publish returned 200 but the API stayed Created" — not as a config problem, which is what makes it expensive.

This PR already establishes the right pattern twice: ELK analytics and the mail sink each got their **own block** precisely because their config is server-global. Same treatment here — a `WorkflowWsApiStateChangeRunner` plus an `IntegrationV2-WsApiStateChange` block (`thread-count=1`, `initBackend=true`, the `approveWorkflow` overlay) removes the ordering dependency outright.

If the extra container boot isn't worth it, the alternative is a failure-safe per-scenario restore: a `I register the workflow executors for cleanup` step alongside the existing `I register tenant configuration {string} for cleanup`, which is the same net this suite already uses for global tenant-config mutations.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/features/analytics/elk_metrics.feature
```diff
@@ -0,0 +1,91 @@
...
+    Then The response status code should be 200
+    # The event is published asynchronously after the response is returned, so this polls rather than reading once.
+    # One contiguous substring, so ALL FOUR facts are pinned on the SAME line: it is THIS api's event, the proxy
+    # response code it recorded is the 200 asserted above, the destination is the backend the request really
+    # reached, and the creator tenant is THIS row's tenant. MEASURED, not guessed: destination is the production
+    # endpoint URL of create_apim_test_api.json verbatim, INCLUDING its trailing slash (the ENDPOINT_ADDRESS
+    # message-context property is the configured address, unnormalised) — contrast the respond-mediator scenario
+    # below, where the property is absent and the product substitutes the placeholder.
+    And The server log file "apim_metrics.log" should gain a line containing "INFO ELKCounterMetric apimMetrics: apim:response, properties :{\"apiName\":\"{{elkApiName}}\",\"proxyResponseCode\":200,\"destination\":\"http://nodebackend:3001/jaxrs_basic/services/customers/customerservice/\",\"apiCreatorTenantDomain\":\"<tenantDomain>\"" within 60 seconds
```
AnuGayan: 🟡 **minor — this assertion depends on Gson's `HashMap` iteration order**

The single contiguous substring pins `apiName` → `proxyResponseCode` → `destination` → `apiCreatorTenantDomain` in that exact adjacency. I confirmed the mechanism upstream rather than taking it from the comment: `ELKMetricEventBuilder` builds the event as `eventMap = new HashMap<>()` (then re-filters it through `EventMapAttributeFilter`), and `ELKCounterMetric` logs `gson.toJson(event)` — so the serialised field order is bucket order, a function of the key hashes and the table capacity. Add or remove one analytics attribute upstream, the table resizes, the fields reorder, and both scenarios in this file fail as "the log never gained the line" — indistinguishable from "analytics stopped publishing entirely".

Worth noting where legacy drew the line: `APIMAnalyticsTest.EXPECTED_LOG_OUTPUT` pinned only `{"apiName":"APITest","proxyResponseCode":200,` — the same class of dependency, but two fields instead of four. This doubles the exposure.

The same-line guarantee is worth keeping, but not at that price. Either pin the short adjacency legacy relied on (`"apiName":"X","proxyResponseCode":200`) and assert `destination` / `apiCreatorTenantDomain` separately, or add a step taking a **list** of substrings all required on the same appended line — which keeps the stronger claim without pinning field order at all. Same applies to line 86.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/PublisherBaseSteps.java
```diff
@@ -2812,6 +2813,56 @@ public void iDownloadWsdlFromStore(String apiId) throws IOException {
...
+        JSONArray policies = new JSONObject(list.getData()).getJSONArray("list");
+        Assert.assertEquals(policies.length(), 1,
+                "Expected exactly one in-sequence resource policy, got: " + list.getData());
+        JSONObject policy = policies.getJSONObject(0);
+        String content = policy.getString("content");
+        String markerElement = "<property name=\"" + markerValue + "\" value=\"" + markerValue + "\" scope=\"default\"/>";
+        int firstTagEnd = content.indexOf('>');
+        String updatedContent = firstTagEnd >= 0
+                ? content.substring(0, firstTagEnd + 1) + markerElement + content.substring(firstTagEnd + 1)
+                : content + markerElement;
```
AnuGayan: 🔵 **nit — `indexOf('>')` is a naive scan into markup** *(edited — see correction below)*

**Correction to my original comment:** I first framed this as "the marker lands outside the root element". That was wrong, and I checked it: the resource-policy `content` is **not** a rooted document, it is a *fragment* of sibling mediators (see the legacy expected sequence `tests-backend/src/test/resources/artifacts/AM/soap/in-sequence-check-phone-numbers.xml`, which opens with a self-closing `<header …/>` followed by sibling `<property …/>` / `<filter>` elements). Inserting after the first element's `>` therefore yields a perfectly well-formed sibling insertion, and my suggested `lastIndexOf("</")` alternative would have been *worse* — for a fragment ending in `</filter>` it would drop the property **inside** the filter block.

What remains is smaller: `indexOf('>')` scans raw text with no notion of markup, so it lands in the wrong place if the first `>` is not a tag terminator — inside an attribute value (`>` is legal unescaped in XML attributes, and synapse expressions/regexes can carry it), or inside a leading comment that contains one. The result is corrupt XML → a 400 on the PUT or a swallowed marker, and the re-retrieve assertion then reads as a product bug.

Since the content is a fragment, the robust form is also the simplest — prepend, no scanning at all:

```java
String updatedContent = markerElement + content;
```

A property mediator at the head of the sequence is exactly as observable, and it cannot be positioned wrongly.

all-in-one-apim/modules/integration-v2/tests-common/testcontainers/src/main/resources/nodeapps/duplicate-header-backend/controllers/duplicateHeaderController.js
```diff
@@ -26,4 +26,26 @@
...
+        + 'Connection: close\r\n'
+        + '\r\n'
+        + Buffer.byteLength(body).toString(16) + '\r\n' + body + '\r\n0\r\n\r\n';
+
+    const socket = res.socket;
+    // Detach so Node does not append its own (normalised) response onto the same socket.
+    if (typeof res.detachSocket === 'function') {
+        res.detachSocket(socket);
+    }
+    socket.write(raw);
```
AnuGayan, line 55: 🟡 **minor — unguarded `res.socket`**

On an aborted or already-destroyed connection `res.socket` is `null`, so `socket.write(raw)` throws a `TypeError` out of the Express handler. Cheap guard:

```js
const socket = res.socket;
if (!socket || socket.destroyed) {
    return;
}
```

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/utils/Utils.java
```diff
@@ -575,6 +575,23 @@ public static String getResourceEndpointURL(String baseUrl, String resourceType,
...
+    /**
+     * Publisher SOAP-to-REST conversion resource-policy list: {@code /apis/{apiId}/resource-policies?sequenceType=}
+     * (GET). {@code sequenceType} is the one required query param (in|out); returns the {@code {list:[...]}} of
+     * generated in/out sequences. The generic resource GET ({@link #getResourceEndpointURL}) cannot express this
+     * sub-path plus required query param, so it has its own builder.
+     */
+    public static String getApiResourcePoliciesURL(String baseUrl, String apiId, String sequenceType) {
+        return baseUrl + Constants.DEFAULT_APIM_API_DEPLOYER + "apis/" + apiId
+                + "/resource-policies?sequenceType=" + urlEncode(sequenceType);
+    }
```
AnuGayan, line 587: 🟡 **minor — two builders for the same endpoint, 700 lines apart**

This 3-arg overload duplicates the query-string construction of the existing 5-arg `getApiResourcePoliciesURL` at ~line 1314. Have this one delegate (or at minimum co-locate the two) so the path lives in one place.

Also: moving `getApiResourcePolicyByIdURL` up here left a stray double blank line behind at its old site (~1319-1320).

all-in-one-apim/modules/integration-v2/tests-common/testcontainers/src/main/resources/nodeapps/duplicate-header-backend/controllers/digestAuthController.js
```diff
@@ -0,0 +1,91 @@
...
+}
+
+const handleDigestRequest = (req, res) => {
+    const authHeader = req.header('Authorization') || '';
+    if (!authHeader.startsWith('Digest ')) {
+        return challenge(res, 'no digest credentials');
+    }
+
+    const params = parseDigestParams(authHeader.substring('Digest '.length));
+    const password = USERS[params.username];
```
AnuGayan: 🔵 **nit** — unguarded object index: `__proto__` / `constructor` resolve to inherited members rather than `undefined`. Not exploitable here (the hash comparison still fails and the request is challenged), but `security/detect-object-injection`-style rules flag it. Either make `USERS` a `Map`, or:

```js
const password = Object.prototype.hasOwnProperty.call(USERS, params.username)
    ? USERS[params.username] : undefined;
```

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/APIInvocationSteps.java
```diff
@@ -1271,4 +1275,71 @@ private static int countSseEvents(String body) {
...
+        TestContext.remove(HTTP_RESPONSE_KEY);
+        String resolvedContext = Utils.resolveContextPlaceholders(context);
+        String token = TestContext.resolve(accessToken).toString();
+        String base = Utils.getBaseGatewayUrl();
+        if (base.endsWith("/")) {
+            base = base.substring(0, base.length() - 1);
+        }
+        String endpointUrl = base + (resolvedContext.startsWith("/") ? "" : "/") + resolvedContext;
+        List<String> expectedValues = Arrays.stream(expectedValuesCsv.split(","))
+                .map(String::trim).sorted().collect(Collectors.toList());
```
AnuGayan, line 1359: 🔵 **nit** — comma-splitting the expected values means a header value that *contains* a comma can't be expressed (`Set-Cookie: …; Expires=Wed, 21 Oct …` is the common case). Fine for the two scenarios using it today; worth either a javadoc line stating the limitation or taking a DataTable instead.

AnuGayan (commented):
Review of the integration-v2 Group 4 parity coverage. **Test-only change; no product code touched.** Overall this is high-quality work — I verified mechanically that:

- all **940** newly added feature steps resolve to existing glue (no undefined steps; no ambiguous or duplicate step expressions across the 42 glue classes),
- all **35** referenced artifact paths exist in the tree,
- all **829** `coverage-tree.md` references land on real `Scenario` / `Scenario Outline` lines (the **893** total is leaf-lines-with-refs, scenarios mapped under more than one capability included),
- every added `Examples` table is well-formed, with no placeholder missing from its header and no unused column,
- no sleeps and no TODOs were introduced,
- the new `snakeyaml` declaration resolves against the parent's managed `2.2` (which rejects global tags by default, so `new Yaml()` needs no `SafeConstructor` here),
- both new runners are wired into `testng-v2.xml`, no runner globs a feature directory that would pick these features up in a block without their overlay, and `initBackend=true` really is present on `IntegrationV2-ApprovalWorkflow` where the new comment claims it is required.

**Verdict: approve with comments.** Nothing here blocks. One test-ordering dependency (the `workflow_ws_api_state_change` flip) is worth fixing before merge because its failure mode is misleading rather than loud; the rest are minor robustness/maintainability items.

**Particularly good:** the `java.net.http` → Apache `CloseableHttpClient` swap in `ApplicationBaseSteps` fixes a real client leak (every OAuth leg is now in try-with-resources) with an accurate rationale for dropping hostname verification; the JWKS signature assertion closes a genuine hole where a garbage signature satisfied every existing backend-JWT check, and pinning the *distinct kid set* is the stronger form; the mail-count expectation is derived from "one message per published+subscribed predecessor" with arrival-then-settle so a duplicate is caught rather than missed; and the block-isolation reasoning in `testng-v2.xml` is load-bearing documentation, not decoration.

AnuGayan (commented):
## Verification pass — accuracy of this PR's claims (and a correction to my own earlier comment)

Follow-up to my [previous review](https://github.com/wso2/product-apim/pull/14299#pullrequestreview-5006559567). This PR's comments make a lot of load-bearing factual assertions ("MEASURED, not guessed", hardcoded product paths, port-publication reasoning, cucumber ordering), so I checked the ones that are statically checkable against primary sources rather than taking them on faith.

**Verdict: content verified — I found no accuracy defect in the PR.** The one defect this pass surfaced was in *my* review, and it is corrected below.

### Correction to my own review

My comment on `PublisherBaseSteps.java:2857-2860` was **wrong in its rationale**, and the fix I suggested would have made things worse. I claimed the marker "lands outside the root element". It does not: the resource-policy `content` is not a rooted document but a **fragment of sibling mediators** — see the legacy expected sequence `tests-backend/src/test/resources/artifacts/AM/soap/in-sequence-check-phone-numbers.xml`, which opens with a self-closing `<header …/>` followed by sibling `<property …/>` / `<filter>` elements. Inserting after the first element's `>` is therefore well-formed, and my suggested `lastIndexOf("</")` alternative would have dropped the property **inside** the trailing `</filter>` block.

What survives is a much smaller nit (`indexOf('>')` scans raw text and can land inside an attribute value or a leading comment; `markerElement + content` avoids scanning entirely). [That comment](https://github.com/wso2/product-apim/pull/14299#discussion_r3842366633) has been rewritten with an explicit correction note and dropped to a 🔵 nit — apologies for the noise.

### PR claims verified against primary sources

| Claim | Checked against | Result |
|---|---|---|
| `APIStateChangeWSWorkflowExecutor` POSTs to `<serviceEndpoint>` + a hardcoded `/runtime/process-instances` | `carbon-apimgt` `APIStateChangeWSWorkflowExecutor.java:72,188` | ✅ exact — and `GET …?businessKey=` (:305) / `DELETE …/{processId}` (:327) match the BPMN double's three routes, so the dual mount in `server.js` and the bare `nodebackend:3004` `serviceEndpoint` are both right |
| The executor DCR-mints `clientId`/`clientSecret` and sends `scope` | same file :400-411, :479; `apim:api_workflow` in `WorkflowConstants` | ✅ |
| "Array order is not honoured — cucumber-testng sorts by filename" | `cucumber-core` 7.0.0: `FeaturePathFeatureSupplier$FeatureBuilder` calls `Comparator.comparing(…)` + `List.sort(…)` | ✅ confirmed — which is also what makes my 🟠 finding on `WorkflowApprovalRunner` real rather than hypothetical |
| The ELK properties object is a `HashMap` serialised by Gson | `ELKMetricEventBuilder` (`eventMap = new HashMap<>()`), `ELKCounterMetric` (`gson.toJson(event)`, `"apimMetrics: … properties :"`) | ✅ — the `INFO ELKCounterMetric` prefix also matches legacy's own constant |
| 3025 is deliberately unpublished; 3005 is published; a listed-but-silent port stalls boot | `NodeAppServer.java:35-36` (3000-3019, 3021-3024 — no 3020, no 3025) + `Wait.forListeningPort()` | ✅ including the "3020 mcp-server precedent" |
| The new npm deps will actually ship | `nodeapps/Dockerfile`: `duplicate-header-backend` appears in both the `COPY package*.json` list and the `npm install` loop | ✅ — and `smtp-server ^3.13.4` → 3.19.3, `mailparser ^3.6.9` → 3.9.15 both resolve |
| CLAUDE.md: "`node-soap-service` sits in the tree but is absent from the `COPY` list" | Dockerfile — zero occurrences | ✅ |
| `snakeyaml` version is inherited from the parent's `dependencyManagement` | `all-in-one-apim/pom.xml:934` + `version.snake.yaml = 2.2` | ✅ (and 2.2's `LoaderOptions()` defaults to `UnTrustedTagInspector`, whose `isGlobalTagAllowed` returns `false` — so `new Yaml()` needs no `SafeConstructor` here) |
| Legacy `NotificationTestCase` / `APIStateChangeWorkflowTestCase` were vacuous / non-null-only | `APIStateChangeWorkflowTestCase.java:165-205` — `assertNotNull` on `businessKey`/`scope`, values merely forwarded | ✅ the characterisation is fair |
| `initBackend` is present where the new comments say it is required | `testng-v2.xml:760` | ✅ |

### No regression from the two shared-step semantic changes

- **`iResetApplicationThrottlePolicy`** now goes through `Identity.apiUsername(Identity.resolveActor(owner))`, which *throws* on an unprovisioned key. All five call sites (three new, two pre-existing on master) pass actor references (`<actor>` / `<ownerActor>`), never literal usernames — checked on the branch and by code search on master. Safe.
- **`iSetJsonFieldFromFileInPayload`** (object-or-array) and **`/reflect-headers`** (`router.get` → `router.all` + catch-all text parser) are both additive: a bodyless GET leaves `req.body = {}`, so the reflected payload stays byte-for-byte what it was.

### Not statically verifiable (flagging, not objecting)

The exact-value pins in `workflow_ws_api_state_change.feature` — `CREATED`, `admin-AT-tenant1.com`, the `callbackUrl` — cannot be confirmed from source: legacy only asserted non-null, and `setApiCurrentState`'s callers aren't reachable via code search. They are genuinely runtime-measured. That's acceptable, because a mis-measurement fails the scenario loudly rather than passing vacuously — the right failure direction.

### One addition

3025 is missing from the Dockerfile's `EXPOSE` list as well as from `exposedPorts`. Functionally irrelevant (`EXPOSE` is documentation-only and in-network traffic doesn't need it), but that line is the image's port inventory — worth listing it with an "SMTP, in-network only" note so the next reader doesn't conclude the sink isn't there.

### Closing note

The prose in this PR is unusually trustworthy — every independently checkable claim held, down to the hardcoded BPMN resource path, the port-publication reasoning and the cucumber ordering rule. My earlier verdict stands: **approve with comments**, with the `workflow_ws_api_state_change` ordering dependency the one item I'd fix before merge, because its failure mode is misleading rather than loud.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/features/gateway/soap_invocation.feature
```diff
@@ -110,6 +174,296 @@ Feature: Gateway SOAP API Invocation
...
+    When I update the "apis" resource "soApiId" and "soPayload" with configuration type "scopes" and value:
+      """
+      [{"shared":true,"scope":{"name":"s2rScopeEnf","displayName":"s2rScopeEnf","description":"s2r operation scope enforcement","bindings":["admin"]}}]
+      """
+    Then The response status code should be 200
+    When I retrieve the "apis" resource with id "soApiId"
+    And I put the response payload in context as "soPayload"
+    When I update the "apis" resource "soApiId" and "soPayload" with configuration type "operations" and value:
+      """
+      [{"target":"/sayHello","verb":"POST","authType":"Application & Application User","throttlingPolicy":"Unlimited","scopes":["s2rScopeEnf"],"operationPolicies":{"request":[],"response":[],"fault":[]}}]
```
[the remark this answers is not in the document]
nimsara66: @coderabbitai,

Already addressed. s2rScopeEnf is now only a base name — I create a new shared scope as uniquifies it via Names.unique, the scenario captures the created name (I extract response field "name" and store it as "s2rScopeName"), and the scopes config, operations config and token request all reference {{s2rScopeName}}.

Applied uniformly across all 12 shared-scope sites, not just this one, since every other resource class in the suite already uniquifies and scopes were the sole outlier.

all-in-one-apim/modules/integration-v2/tests-common/testcontainers/src/main/resources/nodeapps/duplicate-header-backend/controllers/duplicateHeaderController.js
nimsara66: Guard applied — placed before detachSocket(), since detaching is what nulls res.socket.

One correction: the stated mechanism didn't reproduce. On an aborted request res.socket is a live object with destroyed=true (not null), and since the handler captures it into a local before detaching, socket.write() neither threw nor emitted an uncaught error — 3 aborted requests left zero TypeError/ERR_STREAM_DESTROYED in the app log and the route still served 200.

Kept anyway as an intent guard (it stops the handler detaching and writing to a dead connection), with the measured behaviour recorded in the comment rather than the null-deref claim.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/utils/Utils.java
nimsara66: Deduplicated — though the delegation had to go the other way: the 3-arg has no resourcePath/verb, so the 5-arg now narrows it (getApiResourcePoliciesURL(baseUrl, apiId, sequenceType) + "&resourcePath=…&verb=…"), leaving the path spelled once. This also restores the file's existing convention: every other overload family here (getAPIComment(s), getDevportalApiListURL, getDevportalMcpServerListURL) already delegates — this was the sole deviation.

Stray blank line removed. It was from deleting a duplicate getApiResourcePolicyByIdURL during the rebase, not from moving one; the two remaining double-blanks (1707, 2366) predate the branch.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/APIInvocationSteps.java
nimsara66: Addressed — added a LIMITATION paragraph to the step's javadoc (APIInvocationSteps.java:1338) stating that expectedValuesCsv is comma-split, so a value containing a comma (dated Set-Cookie) can't be expressed and would fail the count assertion.

Neither scenario using the step needs one today, so the signature is unchanged.

It also records the fix if that changes: take the markers as a DataTable, per LoggingSteps#logFileShouldGainLineContainingAll, rather than inventing a second delimiter.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/testng-v2.xml
```diff
@@ -125,6 +125,9 @@
         <parameter name="blockLabel" value="keymanager"/>
         <parameter name="initTenantUsers" value="true"/>
         <parameter name="initBackend" value="true"/>
+        <!-- Purely additive SECONDARY.COM store (own H2 DB, no primary user/role/config change) — seeds the store
+             actors the token_issuance store-principal token arcs act as (§12). -->
+        <parameter name="initSecondaryUserStore" value="true"/>
```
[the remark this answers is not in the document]
nimsara66: @coderabbitai,

Addressed by the finding's second option: the 3 SECONDARY.COM store-principal scenarios were extracted to token_issuance_store_principal.feature with its own runner, registered in IntegrationV2-KeyManager only. token_issuance.feature now has 0 store refs, so it runs safely in both blocks.

We deliberately did not add initSecondaryUserStore to EmailUserName: measured, it takes the failures 6 → 3, not 6 → 0. EnableEmailUserName splits on the last @ only when a name carries two or more, so SECONDARY.COM/subscriberUser1@tenant1.com resolves to carbon.super and 401s — the rule basic_auth_email_username.feature exists to pin.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/APIInvocationSteps.java
```diff
@@ -1271,4 +1308,78 @@ private static int countSseEvents(String body) {
...
+
+        Map<String, String> requestHeaders = Identity.bearerHeaders(token);
+        AtomicInteger lastStatus = new AtomicInteger(-1);
+        // Retry only ROUTABILITY (a non-200 while the gateway warms up); once a 200 arrives its headers ARE the
+        // answer and are asserted below — a collapsed header set must fail, not be polled away.
+        List<String> observed = Utils.retryUntil(timeoutSeconds * 1000L, () -> {
+            SimpleHTTPClient.MultiValuedHeaderResult result = SimpleHTTPClient.getInstance()
+                    .doRequestCollectingHeader(httpMethod, endpointUrl, requestHeaders, headerName);
```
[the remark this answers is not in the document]
nimsara66: @coderabbitai,

Skipping this one. doRequestCollectingHeader returns MultiValuedHeaderResult, not HttpResponse — and HttpResponse can't represent repeated headers, which is the step's entire subject (detecting a collapsed or multiplied duplicate set). So it can't be published as httpResponse: wrong type, and it would mislead the generic assertion steps.

execute(...) is clear → call → set(httpResponse) → return. Strip the publish half and it collapses to exactly the TestContext.remove + raw-client call already there, while making the funnel publish for some return types and not others — which weakens the stale-response discipline it exists to enforce. CLAUDE.md §7 makes the same point about a "don't publish" flag on Requests.

The step already does the one thing execute would add (clears httpResponse before the call, so a throw leaves it absent rather than stale).

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/PublisherBaseSteps.java
```diff
@@ -279,8 +280,40 @@ public void iCreateResourceRevision(String resourceType, String resourceId, Stri
...
+     */
+    private void awaitRevisionReadable(String resourceType, String resourceId, String revisionId)
+            throws IOException, InterruptedException {
+
+        String url = Utils.getRevisionURL(Utils.getBaseUrl(), resourceType, resourceId);
+        Map<String, String> headers = Identity.publisherHeaders();
+        HttpResponse last = Utils.retryUntil(Constants.RUNTIME_PROPAGATION_TIMEOUT,
+                () -> SimpleHTTPClient.getInstance().doGet(url, headers),
+                response -> response != null && response.getResponseCode() == 200
+                        && response.getData() != null && response.getData().contains(revisionId));
```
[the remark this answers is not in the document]
nimsara66: @coderabbitai,

Skipping both. These are §7-sanctioned intermediate reads: consumed locally within the step, not the asserted response. A "non-publishing Requests GET" collapses to exactly the raw-client call already there — §7 calls this out explicitly for Requests.

Concretely harmful, not just stylistic: iCreateResourceRevision publishes the create's 201, and 6 scenarios in api_revisions.feature (:28, :57, :78, :131, :201, :239) assert it. The revision poll runs after the create in the same step, so publishing its GET 200 would clobber the 201 and fail all six.

:2903 is the same shape and already documents it inline. No change made.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/features/gateway/mediation_policies.feature
```diff
@@ -232,6 +266,114 @@ Feature: Gateway Mediation Policies
...
+    And I have created an api from "artifacts/payloads/create_apim_reflect_api.json" as "baseSecretApiId" and deployed it
+    When I publish the "apis" resource with id "baseSecretApiId"
+    Then The lifecycle status of API "baseSecretApiId" should be "Published"
+    When I retrieve the "apis" resource with id "baseSecretApiId"
+    And I extract response field "context" and store it as "baseSecretContext"
+    When I have set up application with keys, subscribed to API "baseSecretApiId", and obtained access token for "baseSecretSubId"
+    Then The response status code should be 200
+
+    # BASELINE: with no policy attached, neither the secret value nor the injected header names reach the backend.
+    When I invoke the API at gateway context "{{baseSecretContext}}/1.0.0/reflect-headers" with method "GET" using access token "generatedAccessToken" and payload "" until response status code becomes 200 within 60 seconds
```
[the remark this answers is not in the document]
nimsara66: @coderabbitai,

Fixed — all 8 now carry should be live on the gateway, redeploying if propagation is lost before their first gateway read. The 3 soap_invocation sites were already correct; the 5 in mediation_policies and sandbox_gateway had the gate positioned after the first read (before a later second deploy), leaving the baseline invocation unprotected.

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/java/org/wso2/am/integration/cucumbertests/stepdefinitions/PublisherBaseSteps.java
```diff
@@ -3101,6 +3232,92 @@ public void theExportedPolicyArchiveShouldContain(String archivePathKey, String
...
+        Assert.assertTrue(exportedSpec instanceof Map && ((Map<?, ?>) exportedSpec).get("data") instanceof Map,
+                "Exported spec has no 'data' object: " + specContent);
+        Object exportedData = ((Map<?, ?>) exportedSpec).get("data");
+        assertSpecContentContainsSource(sourceData, exportedData, "data");
+
+        String sourceSynapse = normalizeSynapse(Utils.readClasspathResource(sourceSynapseResource));
+        String exportedSynapse = normalizeSynapse(
+                new String(Files.readAllBytes(synapseFile.toPath()), StandardCharsets.UTF_8));
+        Assert.assertEquals(exportedSynapse, sourceSynapse,
+                "Exported synapse template content does not match the source " + sourceSynapseResource);
```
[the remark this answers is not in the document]
nimsara66: @coderabbitai,

Fixed — PublisherBaseSteps:3253 now compares raw source against raw exported content, UTF-8 decoding and the mismatch message unchanged. The now-unused normalizeSynapse helper is deleted.

nimsara66:
@coderabbitai,

Please approve.

AnuGayan (approved):

PR #14301, opened 2026-08-28
Add integration-v2 coverage for AI API non-hybrid gateway deployment and MCP server cross-environment import

Oshanath:
Integration-v2 coverage for two fixes, one scenario each. Both fail on an unfixed server and pass on a fixed one.

### AI API deployment to a non-hybrid gateway

An AI API keeps its endpoints as named endpoints, outside the endpoint configuration, so clearing that configuration leaves the API with no production endpoint there while a named production endpoint remains. The check that refuses a deployment when the API carries no endpoint of the stage the gateway serves read the endpoint configuration alone and refused such an API, which then never reached the gateway.

That check is only evaluated for a gateway serving **one** stage. Every existing AI scenario deploys to `Default`, which is hybrid, so their coverage stays green whether or not the deployment is refused — hence a new block whose overlay redefines `Default` as production-only. It cannot be co-hosted: a production-only `Default` stops every API carrying a sandbox endpoint alone from being deployable. Adding an environment instead of redefining `Default` does not work either — an environment created through the admin API is registered in the control plane but is not served by this container's gateway (`gateway_labels` names `Default` alone), so an API deployed only there never routes.

The scenario asserts both halves of the premise before deploying, so a premise that did not hold is reported as such instead of surfacing later as a routing failure, and then asserts the subject by invoking through the gateway, since a refused deployment never routes.

### MCP server cross-environment import

An MCP server built on an existing API records that API's id, and that id belongs to the environment the archive was exported from. Importing it elsewhere failed even when the same API was present, as a generic server error. The reference is resolved by the name and the version the archive also records.

There was no import or export coverage of MCP servers at all. A second environment is stood in for by rewriting the recorded id in the exported archive, which meets the same condition an import into a genuinely separate environment meets without a second server to run. The rewrite fails when it matched nothing, so an archive whose shape changed cannot leave the scenario importing the id it was exported with and passing while proving nothing.

The **import alone** is performed as the admin actor: it accepts `apim:mcp_server_import_export` (mapped to `admin,Internal/devops`) or `apim:admin`, and the publisher token carries neither. No token's scope list is widened — the export is content with `apim:mcp_server_view`, which the publisher token already holds.

### Notes

- `Utils.rewriteZipEntry` joins the existing archive helpers (`zipEntryText` / `unzip` / `zipDirectory`); it cannot be expressed as the last two, which keep only an archive's top level.
- The coverage tree is regenerated: 835 covered, 0 unmapped/invalid, 0 not-executed.

### Testing

Both scenarios run locally against a `4.7.0-SNAPSHOT` image carrying the fixes — `Tests run: 2, Failures: 0, Errors: 0`, clean teardown with no leak warnings.


all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/features/gateway/ai_api_non_hybrid_gateway.feature
```diff
@@ -0,0 +1,67 @@
...
+    When I put the following JSON payload in context as "emptiedEndpointConfig"
+    """
+    {"endpoint_type":"http"}
+    """
+    And I retrieve the "apis" resource with id "epRemovalApiId"
+    And I put the response payload in context as "epRemovalApiPayloadWithPrimary"
+    And I update the "apis" resource "epRemovalApiId" and "epRemovalApiPayloadWithPrimary" with configuration type "endpointConfig" and value:
+    """
+    emptiedEndpointConfig
+    """
```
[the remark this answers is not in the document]
Oshanath: iUpdateTheResourceWithConfigurationTypeAndValue special-cases configType == "endpointConfig" and treats the docstring as a context key, not a value: it calls TestContext.resolve(configValue) and only then expands {{…}} inside the payload it fetched. Passing {{emptiedEndpointConfig}} would look up a key literally named {{emptiedEndpointConfig}}, which TestContext.resolve throws IllegalArgumentException for. The bare key is the established convention for this step across the suite — versioning.feature:143 (verUpdEndpoint), api_config.feature:564 (epubNewEndpoint), api_product_invocation.feature:124 (epV2Endpoint), rest_invocation.feature:470 (ceNewEndpoint), default_version_routing.feature:55 (dvV2Endpoint) all do the same. The review's stated consequence — "the update can fail or store a wrong-shaped endpoint configuration, so the later absence check does not prove the intended regression premise" — is also directly contradicted by the run: Then The response should not contain "production_endpoints" passes, which it could not if the literal string had been stored as the endpoint configuration.

Oshanath:
@coderabbitai approve

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/features/gateway/ai_api_non_hybrid_gateway.feature
```diff
@@ -0,0 +1,67 @@
...
+  An AI API keeps its endpoints as named endpoints, outside the endpoint configuration. Clearing that configuration
+  therefore leaves it carrying no production endpoint while a named production endpoint remains. The check that
+  refuses a deployment when the API carries no endpoint of the stage the gateway serves read the endpoint
+  configuration alone, so it refused such an API and it never reached the gateway.
+
+  That check is only evaluated for a gateway serving ONE stage, which is why this block redefines Default as
+  production-only (see the tomlExtraOverlayPath on its test block). The AI coverage in
+  gateway/ai_api_invocation.feature deploys to a hybrid Default and so stays green either way.
+
+  @cap:gateway @feat:ai-invocation @rule:non-hybrid-deployment @type:regression @dep:admin @dep:publisher
```
nimsara66, line 13: Why is this not running for both tenants?

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/features/publisher/mcp_servers.feature
```diff
@@ -286,3 +286,39 @@ Feature: MCP Server authoring (publisher plane)
...
+  #
+  # A second environment is stood in for by rewriting the recorded id to one that belongs to no API here — the same
+  # condition an import into a genuinely separate environment meets, without a second server to run.
+  #
+  # The IMPORT alone is performed as the admin actor: it accepts apim:mcp_server_import_export — which
+  # tenant-conf.json maps to admin,Internal/devops — or apim:admin, and the publisher token carries neither, so it
+  # is refused as unauthenticated. Everything else here, the export included, runs on the publisher token. This is
+  # also why the scenario is not an outline over publisherUser like its siblings.
+  @cap:publisher @feat:mcp-servers @rule:cross-environment-import @type:regression
+  Scenario: An MCP server whose archive records a reference API id from another environment is imported
```
nimsara66, line 302: Why is this not running for both tenants?

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/features/gateway/ai_api_non_hybrid_gateway.feature
Oshanath: Done

all-in-one-apim/modules/integration-v2/tests-integration/cucumber-tests/src/test/resources/features/publisher/mcp_servers.feature
Oshanath: Done

